PR #128, opened 2020-07-13
Fix issue with newarray panicing on imported types.

### LINKED ISSUE #120: Compiler panics on new array of imported type
An expression like `newarray<type>(expr)` causes the compiler to panic if `type` is an imported type.  This is inconsistent with the Mini language spec, which says the expression should create an array of uninitialized values.

RGates94:
Fixes #120

### DIFF
--- a/src/compile/ast.rs
+++ b/src/compile/ast.rs
@@ -229,61 +229,53 @@ impl Type {
         }
     }
 
-    pub fn default_value(&self) -> Value {
+    pub fn default_value(&self) -> Result<Value, String> {
         match self {
-            Type::Void => {
-                panic!("tried to get default value for void type");
-            }
+            Type::Void => Err("tried to get default value for void type".to_string()),
             Type::Uint | Type::Int | Type::Bytes32 | Type::EthAddress | Type::Bool => {
-                Value::Int(Uint256::zero())
+                Ok(Value::Int(Uint256::zero()))
             }
             Type::Tuple(tvec) => {
                 let mut default_tup = Vec::new();
                 for t in tvec {
-                    default_tup.push(t.default_value());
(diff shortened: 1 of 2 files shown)


src/compile/ast.rs
```diff
@@ -229,61 +229,53 @@ impl Type {
...
-                value_from_field_list(vals)
+                Ok(value_from_field_list(vals))
             }
             Type::Map(_key, _val) => {
                 // an unusable dummy value -- application will panic if it accesses this
-                Value::none()
-            }
-            Type::Named(_) => {
-                panic!("tried to get default value for a named type");
+                Ok(Value::none())
```
edfelten, line 269: The default value for a Map should be Tuple(0).  That corresponds to a None value in the language, which is the representation for an empty Map.

src/compile/ast.rs
```diff
@@ -229,61 +229,53 @@ impl Type {
...
-            Type::Imported(_) => {
-                panic!("tried to get default value for an imported type");
-            }
-            Type::Any => Value::none(),
-            Type::Every => {
-                panic!("tried to get default value for the every type");
+                Err("tried to get default value for a function type".to_string())
             }
-            Type::Option(_) => Value::new_tuple(vec![Value::Int(Uint256::zero())]),
+            Type::Imported(_) => Err("tried to get default value for an imported type".to_string()),
```
edfelten, line 275: This is better than having a panic, but it still could be better, because there is now no way to make an array of imported type.   It would be better to have a version of `newarray` that takes a base value. That way, if you have one instance of a type, you can safely create and initialize an array of that type.

RGates94:
I'll be resolving this with a different approach, so I'm going to close this PR.

PR #130, opened 2020-07-14
Batched messages and ecrecover

edfelten:
* Implement ecrecover AVM instruction
* Process aggregated batches of messages on the server side
* Generate signed messages on the client side, and verify them (in messages from batches) in ArbOS
* Fix bug in ArbSys precompile
* Add ecrecover precompiled contract at address 1
* Refactor handling of precompiles


edfelten:
Closing this because files are corrupted.

PR #134, opened 2020-07-15
Translate illegal instructions more safely

edfelten:
This modifies the EVM->AVM translator to do something reasonable for all invalid or unsupported EVM instructions.

### DIFF
--- a/arb_os/codeSegment.mini
+++ b/arb_os/codeSegment.mini
@@ -272,7 +272,7 @@ public impure func translateEvmCodeSegment(
                         if (opcode == 0x43) { // NUMBER
                             return pushEvmInsnCall(evmOp_number, restOfCode, evmJumpTable);
                         } elseif (opcode == 0x44) { // DIFFICULTY
-                             return Some((restOfCode, evmJumpTable));  // no-op
+                             return Some((pushValue(2500000000000000, restOfCode), evmJumpTable));
                         }
                     }
                 } else {
--- a/arb_os/codeSegment.mini
+++ b/arb_os/codeSegment.mini
@@ -272,8 +272,7 @@ public impure func translateEvmCodeSegment(
                         if (opcode == 0x43) { // NUMBER
                             return pushEvmInsnCall(evmOp_number, restOfCode, evmJumpTable);
                         } elseif (opcode == 0x44) { // DIFFICULTY
-                             return Some((restOfCode, evmJumpTable));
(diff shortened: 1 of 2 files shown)


arb_os/codeSegment.mini
```diff
@@ -276,7 +272,7 @@ public impure func translateEvmCodeSegment(
                         if (opcode == 0x43) { // NUMBER
                             return pushEvmInsnCall(evmOp_number, restOfCode, evmJumpTable);
                         } elseif (opcode == 0x44) { // DIFFICULTY
-                            return Some((pushValue(0, restOfCode), evmJumpTable));
+                             return Some((restOfCode, evmJumpTable));
```
hkalodner, line 275: What does this do?

arb_os/codeSegment.mini
```diff
@@ -342,7 +338,7 @@ public impure func translateEvmCodeSegment(
                     }
                 } else {
                     if (opcode == 0x58) {  // GETPC
-                        return Some((pushValue(restOfCode, restOfCode), evmJumpTable));
+                        return Some((pushValue(bytestream_bytesReadSoFar(bs), restOfCode), evmJumpTable));
```
hkalodner, line 341: This is tricky so if we're looking to make sure we're compatible it would be nice to have a test. One simple option would be to write a solidity contract which has a function which returns a value based on `GETPC`, run it in geth/ganache in order to figure out what the correct result is, and then run it in ArbOS comparing against that result.

arb_os/codeSegment.mini
edfelten: It's a no-op.

arb_os/codeSegment.mini
edfelten: Hmm, I think there's an off-by-one error in the code you reviewed.  Which is now fixed.

Earlier today I added an issue to do this kind of testing for EVM->AVM translation more generally.

arb_os/codeSegment.mini
hkalodner: The difficulty opcode is expected to push an integer to the stack. We either need to reject code that includes the difficulty opcode, or have the opcode push some integer

arb_os/codeSegment.mini
hkalodner: Sounds good

arb_os/codeSegment.mini
edfelten: Hmm, somehow I misread the spec.  Fixing this to push an integer reflecting a recent difficulty level.

PR #135, opened 2020-07-15
Batched messages and ecrecover

edfelten:
* Implement ecrecover AVM instruction
* Process aggregated batches of txs in ArbOS
* Generate signed txs on the client side, and verify them (in txs from batches) in ArbOS
* Fix bug in ArbSys precompile
* Add ecrecover precompiled contract at address 1
* Refactor handling of precompiles
* Remove erroneous tests


arb_os/precompiles.mini
```diff
@@ -0,0 +1,76 @@
...
+    acctStore = accountStore_createBuiltinContract(
+        acctStore,
+        address(100),
+        arbsys_txcall,
+    )?;
+    return Some(acctStore);
+}
+
+public impure func precompile_0x01() {  // ecrecover
+    // don't bother checking the first four bytes of calldata -- this is the only method in the contract
```
hkalodner, line 57: From the yellow paper:

h = d[0..31]
v = d[32..63]
r = d[64..95]
s = d[96..127]

src/evm/mod.rs
```diff
@@ -178,11 +179,14 @@ pub fn evm_xcontract_call_using_batch(
 ) -> Result<bool, ethabi::Error> {
     use std::convert::TryFrom;
     let rt_env = RuntimeEnvironment::new(Uint256::from_usize(1111));
+
+    let wallet = rt_env.new_wallet();
+    let addr_h160: Address = wallet.private_key().into();
+    let my_addr = Uint256::from_bytes(addr_h160.as_bytes());
```
hkalodner, line 185: Is wallet of type `ethers_signers::Wallet`? If so it seems like we should just use `wallet.address()` here

edfelten: Switched to using `wallet.address()`.  But I still need to convert it to a Uint256.

arb_os/precompiles.mini
edfelten: Changed to match the yellow paper.

hkalodner: Sorry, to be clear I think that based on the spec it looks like it starts from an offset of 0 rather than 4.

edfelten: Fixed. Thanks for pointing this out.

edfelten:
Last change was trivial, and Harry apparently approved merge, so I'm merging this.

hkalodner:
+1 😁

PR #136, opened 2020-07-16
Switch to new L2 message/batch format

edfelten:
Switch message signing and batching to new message format.  Updates ArbOS batch and signature handling, and updates client-side message generator to match.


edfelten:
Going to refactor this based on Slack conversation with Harry.

PR #140, opened 2020-07-16
Rename incoming request object type

edfelten:
This PR just renames a data type. The new name is more intuitive given how the code has changed over time.


edfelten:
Merging this without review, because it only renames things, doesn't change any functionality.

PR #141, opened 2020-07-17
Input queue support for sequencer

edfelten:
This PR adds ArbOS support for queueing incoming messages when using an optional sequencer.  In support of that, it also adds support for parsing the initialization message to extract chain parameters and initialization options.

The sequencer is associated with a parameter `blocksOfDelay`.  Messages from addresses other than the sequencer will be delayed for `blocksOfDelay` L1 blocks, or until the sequencer releases them, whichever comes first. 

A new L2 message type, "sequencer batch", allows the sequencer to release delayed messages that arrived up to some block number specified by the sequencer.  A sequencer batch message also delivers a batch of requests that are to be processed after the released messages.

Every delayed message is released by some other message, either by the first message of any type that arrives at or after `delayedMessage.blockNumber + blocksOfDelay`, or by an earlier sequencer batch message.  If a message is released because its block number is explicitly released by a sequencer batch message, the released message is re-tagged with the block number and timestamp of the message that released it.  If a message is released because `blocksOfDelay` blocks elapsed, the message is re-tagged with a block number equal to its original block number plus `blocksOfDelay`, and with a timestamp equal to that block's timestamp if ArbOS knows that timestamp (otherwise, an interpolated estimate of the block's timestamp).

Released messages are processed *before* all other messages with the same block number.  (This implies that a delayed message will always execute before the message that releases it.)


edfelten:
Closing this as superseded by #244 .

PR #144, opened 2020-07-17
Calculate requestId for non-signed tx

edfelten:
ArbOS now calculates a requestId for non-signed txs (aka L2 messages of subtype 0). This is put into the requestId field of the log item reporting the tx result.

Formula:
hash2(
    uint32(sender), 
    hash2(
        uint32(chainId), 
        hashBytestack(L2 message as bytestack)
    )
)

### DIFF
--- a/arb_os/signedTx.mini
+++ b/arb_os/signedTx.mini
@@ -66,11 +66,11 @@ public impure func translateSignedTx(req: MessageFromL1) -> option<MessageFromL1
     );
 
    let rlpEncoded = rlp_encodeMessageInfo(
-        bytearray_get256(extractedL2data, 2*32),             // sequenceNum,
-        bytearray_get256(extractedL2data, 1*32),             // gasPriceBid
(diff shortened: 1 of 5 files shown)


arb_os/signedTx.mini
```diff
@@ -87,22 +116,20 @@ impure func recoverSigner(msgData: ByteArray, preSigSize: uint) -> option<addres
...
+    return keccak256(rlpEncoded, 0, bytearray_size(rlpEncoded));
+}
 
-    if (signer == address(0)) {
-        return None;
-    } else {
-        return Some(signer);
-    }
+public func hashL2messageInfo(l2msg: MarshalledBytes, chainId: uint, sender: address) -> bytes32 {
+    asm(300,) { debugprint };
```
hkalodner, line 127: I think this debugprint should be removed. Other than that is PR is solid and I have my message batching tested against it and working

PR #146, opened 2020-07-18
Codegen refactor

RGates94:
Refactors codegen to simplify control flow and make implementations of future features including modules and inlining easier.


src/compile/codegen.rs
```diff
@@ -66,41 +66,45 @@ pub fn mavm_codegen<'a>(
...
 
     let mut label_gen = LabelGenerator::new();
-    let mut code = code_in;
+    let mut code = Vec::new();
+    let mut funcs_code = BTreeMap::new();
     for func in funcs {
         if !func.imported {
-            let (lg, c) = mavm_codegen_func(
+            let id = func.name;
+            let (lg, code) = mavm_codegen_func(
```
edfelten, line 74: The overloading of the variable name `code` might confuse readers here.  It's maybe better to rename one of them.

src/compile/codegen.rs
```diff
@@ -208,42 +202,71 @@ fn add_args_to_locals_table(
...
+}
+
+///Generates code for the provided statement. code represents the code generated previously,
+/// num_locals the maximum number of locals used at any point in the call frame so far, locals is a
+/// map of local variables, label_gen points to the next available locals slot, string_table is used
+/// to get builtins, import_func_map associates each imported function with a label, and
+/// global_var_map maps global variable IDs to their slot number.
+///
+/// If successful the function returns a tuple containing the updated label generator, number of
+/// locals slots used by this statement, a bool that is set to true if the statement causes a
```
edfelten, line 245: It's hard to follow the logic of how locals are handled in this code.  (This has been true since the beginning, but this might be a good time to fix it.). Some statements, like `let`, allocate a new local.  Others, like codeblock statements, will allocate locals within their codegen and then those locals will go out of scope.  

I think it makes sense to refactor by creating a data structure that tracks locals allocation properly, rather than carrying around the separate `num_locals` and locals-map / binding objects and needing to maintain invariants about them everywhere.

edfelten, line 245: I don't think "set to true if the statement causes a return" is exactly accurate.  I think it should say something like "set to true if execution certainly will not continue past this statement (due to a return, panic, or infinite loop on all execution paths in the statement)".

PR #149, opened 2020-07-21
Switch to new tx receipt format

### LINKED ISSUE #148: Make tx receipts more Ethereum-compatible
Additional things that would be useful to put in the transaction receipts:
- _Cumulative gas_: This is a field that ethereum transactions include which is the total amount of gas used in a block after a transaction within that block is finished processing. It would presumably be fairly straightforward to include this info
- _Transaction index_: This says that this transaction was the nth processed in the block
- _Start log index_: This is the total number of evm logs emitted in the block before the given transaction

Those should all be relatively easy. A significantly more involved feature would be a extra block info log emitted for every l1 block which includes a bloom filter for the EVM logs that were output in the block and probably other useful aggregate info about the block.

edfelten:
Add support for the new tx receipt (== AVM log item) format, which restructures the receipt data structure and adds information on: total gas used in L1 block including this tx; index of this tx in L1 block; number of EVM logs so far in this L1 block excluding this tx.

Also refactor the receipt handling code in ArbOS, and in the Rust test harnesses.

Fixes #148

### DIFF
--- a/arb_os/txReceipt.mini
+++ b/arb_os/txReceipt.mini
@@ -38,6 +38,12 @@ type GasUsage = struct {
     gasPriceWei: uint,
 }
 
+type TxResultInfo = struct {
+    returnCode: uint,
+    returnData: MarshalledBytes,
+    evmLogs: EvmLogs,
+}
+
 // result codes:
 //    0    return (success)
 //    1    tx reverted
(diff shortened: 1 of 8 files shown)


arb_os/txReceipt.mini
```diff
@@ -0,0 +1,148 @@
...
+    }
+    let realGasUsage = struct {
+        gasUsed: 0,
+        gasPriceWei: 0,
+    };
+    if let Some(gu) = gasUsage {
+        realGasUsage = gu;
+    }
+
+    let txResultInfo = struct {
```
hkalodner, line 133: I got confused by this struct being called `txResultInfo` and the other `TxResultInfo` struct

edfelten: Resolved this by eliminating the top-level `TxResultInfo` type, which isn't needed anyway.

PR #161, opened 2020-07-26
Block summary logs

edfelten:
* Create a new block-summary log item type. Emit one of them at the end of every L1 block that contains any activity on the chain.
* Handle a new "heartbeat" type of L2 message. This doesn't do any work but just notifies ArbOS that the message's block number and timestamp have been reached. This allows ArbOS to do any time-dependent actions, such as emitting a block summary for a block that has ended.


edfelten:
This is superseded by #178

PR #163, opened 2020-07-27
Switch builtin KVS/map library to more efficient data structure

edfelten:
This data structure is very similar to the one used in the `storageMap` API.  It is somewhat more time- and space-efficient than the previous implementation.


edfelten:
Withdrawing this PR because it has been merged into the Faster Bytearrays PR.

PR #164, opened 2020-07-27
Faster bytearrays

edfelten:
* Rewrite bytearray library to use a custom auto-expanding integer array-tree as its underlying storage, rather than the general-purpose sized array facility used previously.
* Also incorporate KVS data structure change (from other branch and PR)


edfelten:
Closing this, and submitting a PR that also includes benchmarks.

PR #169, opened 2020-07-30
Swap args when compiling EVM BYTE and SIGNEXTEND instructions

edfelten:
EVM and AVM expect their arguments on the stack in reversed order for these two instructions.  Now properly swapping the args when compiling from EVM to AVM.


hkalodner:
This will be superseded by https://github.com/OffchainLabs/mini-compiler/pull/171 and should be closed after that PR is merged.

edfelten:
Closing this because we changed the instruction semantics to eliminate this problem

PR #185, opened 2020-08-10
Support RLP-encoded signed tx in ArbOS

edfelten:
* Change ArbOS to expect signed tx to be submitted in Ethereum-compatible RLP-encoded format
* Add necessary support for this to RLP library code
* Change client runtime support to use the new format
* Improve error reporting in ArbOS
* Refactor message handling pipeline


stdlib/rlp.mini
```diff
@@ -38,6 +39,17 @@ import func hasher_pushN_le(h: Hasher, val: uint, nbytes: uint) -> Hasher;
 import func hasher_pushN_be(h: Hasher, val: uint, nbytes: uint) -> Hasher;
 import func hasher_finish(h: Hasher) -> bytes32;
 
+type SignedTx = struct {
```
hkalodner, line 42: This data structure is duplicated between here and signedtx.mini. There should be a comment on both of them documenting that

arb_os/signedTx.mini
```diff
@@ -15,50 +15,76 @@
...
+    // Given an IncomingRequest that contains an L2 signed tx message, translate the
+    //    L2 message from the RLP-encoded form it arrived in, into the old format which downstream code expects.
+    // Also, check the signature on the tx, and set:
+    //      - the sender field in the request equal to signer, and
+    //      - the requestID field in the request equal to the Ethereum-compatible tx id,
+    // Return the result, or None if message is malformed or the signature is invalid.
+
+    let rlpData = bytearray_unmarshalBytes(req.msgData);
+    rlpData = bytearray_extract(rlpData, 1, bytearray_size(rlpData)-1);
+    let (_, signedTx) = rlp_decodeMessageInfo(bytestream_new(rlpData))?;
```
hkalodner, line 81: I just want to confirm, is this basically just as efficient as creating a new bytestream from rlpData and than skipping the first byte?

arb_os/signedTx.mini
```diff
@@ -15,50 +15,76 @@
...
+
+    let extractedL2data = signedTxToBytes(signedTx);
```
hkalodner, line 83: I know you might refactor this further, but one easy thing to do would be to change this to `signedTxToUnsignedTxBytes` and then we wouldn't need the following code

arb_os/signedTx.mini
edfelten: It's basically the same efficiency.  No copying in either case.

stdlib/rlp.mini
edfelten: Done.

arb_os/signedTx.mini
edfelten: Done.

PR #187, opened 2020-08-10
Profiler updates

RGates94:
Prints a breakdown of the total gas usage of callers to each function as well as a breakdown of gas costs for calls within that function.  The number following the `:` is the gas used by the function excluding gas spent in function calls.


edfelten (approved):
Looks good.  Thanks for the tooling support.

PR #188, opened 2020-08-10
Remove use of Sideload from ArbOS

edfelten:
ArbOS no longer needs to use `sideload` to support the callserver.   After the recent changes to the semantics of inboxes, the callserver can leverage the `inboxpeek` and `inbox` operations in ArbOS to inject messages as needed.


edfelten:
Let's leave sideload in, because it might be useful later, and it does not harm to keep it in the meantime.

PR #190, opened 2020-08-18
Debugprint counters

edfelten:
* Add support for counters that can be manipulated by debug prints. This is useful for debugging and performance measurement.


edfelten:
> This looks fine in terms of behavior, but I think this could be improved architecturally. In particular, rather than having the `counter_store` be part of the `Machine` struct, it would make more sense to have it as a field of the `ProfilerData`.
> 
> This would allow you to leave out the `maybe_machine` argument from `profiler_session`, and would not require or even allow access in code not related to the profiler.

The difficulty with that approach is that `profiler_session` is not available inside the `Machine::run_one_dont_catch_errors` which is where we want to trigger the update of the counter.

RGates94:
> The difficulty with that approach is that `profiler_session` is not available inside the `Machine::run_one_dont_catch_errors` which is where we want to trigger the update of the counter.

To get around this I believe it would be sufficient to move the relevant logic into the `profile_gen` method, where we have equivalent access to the machine's state.

edfelten:
Closing because this is stale and probably not useful anymore.

PR #201, opened 2020-08-25
Lower deposit gas limit, and send L1 refund if deposit fails

### LINKED ISSUE #194: Lower gas limit for ERC20/721 deposits
ERC20 and 721 deposit messages cause ArbOS to call the `adminMint` method of the L2 token contract, with the gas consumed being charged to the ArbOS overhead account. In some scenarios the called code could be untrusted.  To prevent abuse, we need to set the gas limit for these calls to be small.  (At present it's large.)

edfelten:
* Lower the gas limit for ERC-20 and -721 deposits.  This reduces some potential leeching attack scenarios involving use of buddy contracts.
* When an ETH, ERC20, or ERC721 deposit tx fails, automatically send a withdrawal that returns the funds to the sender on L1.

Fixes #194


edfelten:
Closing as superseded by #288

PR #207, opened 2020-09-08
Add support for initialization options

edfelten:
* Add support for options in chain initialization message
* Add initialization option 0, which makes the chain Arbitrum Pro


arb_os/chainParameters.mini
```diff
@@ -37,7 +43,40 @@ public impure func chainParams_gotParamsMessage(sender: address, data: Marshalle
...
+    // on success, return Some(updated stream)
+    // on end-of-stream, return None
+
+    let (ustr, optionId) = bytestream_get64(stream)?;
+    stream = ustr;
+    let (ustr, numPayloadBytes) = bytestream_get64(stream)?;
+    stream = ustr;
+    if (optionId == 0) {   // option: premium chain
+        if (numPayloadBytes != 0) {
+            return None;
```
hkalodner, line 73: Currently this causes the parsing of options to terminate. It might be better to instead just ignore this option and go to the next one

hkalodner (changes requested):

arb_os/chainParameters.mini
edfelten: Changed this to ignore the option and move to the next one if size is unexpected.

hkalodner (approved):
LGTM

PR #214, opened 2020-10-02
Buffer ops

mrsmkl:
I'll try clean up and make a better test for hashing next.


src/run/emulator.rs
```diff
@@ -407,16 +423,32 @@ impl ProfilerData {
...
+        for i in 0..256 {
+            let x = self.instructions[i];
+            total_instructions += x;
+            if x != 0 {
+                if let Some(op) = Opcode::from_number(i) {
+                    println!("{}: {}", op.to_name(), x);
+                }
+            }
+        }
+        println!("Total: {}", total_instructions);
```
RGates94, line 437: I think this part of the profiler changes could go in its own branch.

src/compile/ast.rs
```diff
@@ -220,6 +223,9 @@ impl Type {
             Type::Void => {
                 panic!("tried to get default value for void type");
             }
+            Type::Buffer => {
+                panic!("tried to get default value for buffer type");
```
RGates94, line 227: I think an empty buffer would be a reasonable default value for a buffer type, is there a reason against using this?

src/mavm.rs
```diff
@@ -638,6 +973,83 @@ impl Opcode {
         }
     }
 
+    pub fn to_name(&self) -> &str {
```
RGates94, line 976: I think this could be better as a `Display` impl for `Opcode`.

src/mavm.rs
```diff
@@ -275,12 +275,317 @@ impl fmt::Display for CodePt {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum BufferElem {
+    Leaf(Rc<RefCell<Vec<u8>>>),
+    Node(Rc<RefCell<Vec<Buffer>>>, u8),
+    Sparse(Rc<RefCell<Vec<usize>>>, Rc<RefCell<Vec<u8>>>, u8),
+}
```
RGates94, line 283: I am a bit confused on why we need `RefCell`s here, can you please explain what they are needed for?

src/mavm.rs
```diff
@@ -275,12 +275,317 @@ impl fmt::Display for CodePt {
...
+    if sz == 32 {
+        return normal(zero_hash(32), 32);
+    }
+    return pack(&zero_packed(sz/2));
+}
+
+fn hash_sparse(idx: &[usize], buf: &[u8], sz: usize) -> Packed {
+    if sz == 32 {
+        let mut res = [0u8; 32];
+        for i in 0..idx.len() {
```
RGates94, line 378: I would look into using range based for loops.  There are a number of areas where they could potentially be applied.

src/mavm.rs
```diff
@@ -275,12 +275,317 @@ impl fmt::Display for CodePt {
...
+    let h1 = zero_hash(sz/2);
+    return Uint256::avm_hash2(&h1, &h1);
+}
+
+fn normal(hash: Uint256, sz: usize) -> Packed {
+    return Packed{size: sz, packed: 0, hash: hash};
+}
+
+fn pack(packed: &Packed) -> Packed {
+    return Packed{size: packed.size, packed: packed.packed+1, hash: packed.hash.clone()}
```
RGates94, line 351: I'd recommend making this a `&mut self` method, as it avoids cloning the struct and simplifies the code.

RGates94 (commented):
I would recommend running rustfmt, as it keeps consistent formatting with the rest of the project and saves time.

src/mavm.rs
RGates94: I just noticed that this comment was worded incorrectly.  I meant to say non-range based for loops, i.e. loops using iterators over a collection.

src/run/emulator.rs
```diff
@@ -1654,6 +1697,88 @@ impl Machine {
                             Err(ExecutionError::new("invalid operand to EcPairing instruction", &self.state, None))
                         }
 
+                    }
+                    Opcode::AVMOpcode(AVMOpcode::NewBuffer) => {
+                        let _size = self.stack.pop_uint(&self.state)?;
```
edfelten, line 1702: What is the purpose of the size arg to this instruction?  It seems to be unused.

(In the original ByteArray implementation in the ArbOS library, bytearray_new took a uint arg that was intended as a size hint.)

src/run/emulator.rs
```diff
@@ -1654,6 +1697,88 @@ impl Machine {
...
+                        self.stack.push(Value::new_buffer(vec![0; 256]));
+                        self.incr_pc();
+                        Ok(true)
+                    }
+                    Opcode::AVMOpcode(AVMOpcode::GetBuffer8) => {
+                        let buf = self.stack.pop_buffer(&self.state)?;
+                        let offset = self.stack.pop_usize(&self.state)?;
```
edfelten, line 1709: It's probably better to put the args on the stack in the opposite order, with the offset on top.  That's more efficient because the offset is more likely to be a constant, which can be provided as an immediate value to the GetBuffer8 instruction.

(Same applies to GetBuffer64 and GetBuffer 256.)

src/run/emulator.rs
```diff
@@ -1654,6 +1697,88 @@ impl Machine {
...
+                        let mut res = [0u8; 32];
+                        for i in 0..32 {
+                            res[i] = buf.read_byte(offset+i);
+                        }
+                        self.stack.push_uint(Uint256::from_bytes(&res));
+                        self.incr_pc();
+                        Ok(true)
+                    }
+                    Opcode::AVMOpcode(AVMOpcode::SetBuffer8) => {
+                        let buf = self.stack.pop_buffer(&self.state)?;
```
edfelten, line 1743: Suggest having the order of args on the stack be (top to bottom): offset, val, buf.   Offset is must likely to be constant, so put it on top--and that's consistent with the GetBuffer instructions.  Then put val next to that it's efficient to do a series of SetBuffers in a row by pushing the buffer, then repeating the sequence push val, push offset.

stdlib/bytearray.mini
```diff
@@ -97,38 +69,41 @@ public func bytearray_unmarshalBytes(inBytes: MarshalledBytes) -> ByteArray {
 
     // For efficiency, we'll build an ExpandingIntArray here, then make a ByteArray from it at the end.
```
edfelten, line 70: Remove the obsolete comment here.

stdlib/bytestream.mini
```diff
@@ -4,25 +4,32 @@
...
     capacity: uint,
+    slice: uint,
     currentOffset: uint,
-    cache: option<uint>,
 }
 
 public func bytestream_new(contents: ByteArray) -> ByteStream {
+    // let res = bytearray_copy(contents, 0, bytearray_new(0), 0, bytearray_size(contents));
     return struct {
         contents: contents,
```
edfelten, line 28: Do we need to keep contents here?  It seems like we can eliminate it.

edfelten (commented):
Several suggestions here.  Hashing needs to be fixed to avoid collisions.  The others are non-critical.

edfelten (changes requested):
A few minor changes needed.

edfelten:
I merged in some more tests for the bytearray library, and found some test failures. I fixed those, which required some refactoring of the bytearray library.

stdlib/bytearray.mini
```diff
@@ -78,41 +75,41 @@ public func marshalledBytes_hash(mb: MarshalledBytes) -> bytes32 {
...
     let nbytes = ba.size;
     let nwords = (nbytes+31)/32;
     let words = unsafecast<MarshalledBytesCell>(());
     let i = 0;
     while (i < nwords) {
         words = struct {
+            //first: getbuffer256(ba.contents, 32*i),
             first: bytearray_get256(ba, 32*i),
             rest: words,
         };
```
hkalodner, line 115: Looks like a some unnecessary commented out code added here

stdlib/bytearray.mini
```diff
@@ -2,15 +2,10 @@
 // Copyright 2020, Offchain Labs, Inc. All rights reserved.
 //
 
-use std::expandingIntArray::ExpandingIntArray;
-use std::expandingIntArray::expandingIntArray_new;
-use std::expandingIntArray::expandingIntArray_get;
-use std::expandingIntArray::expandingIntArray_getConsecutive;
-use std::expandingIntArray::expandingIntArray_set;
-use std::expandingIntArray::expandingIntArray_setN;
+use std::random::RandomGenerator;
```
hkalodner, line 5: This import is confusing since it doesn't appear to be used

stdlib/bytearray.mini
```diff
@@ -129,165 +126,139 @@ public func bytearray_size(ba: ByteArray) -> uint {
...
 // bytearray_getByte reads one byte from a ByteArray
 public func bytearray_getByte(ba: ByteArray, offset: uint) -> uint {
     if (offset >= ba.size) {
         return 0;
     }
     offset = offset + ba.sliceOffset;
-    let word = expandingIntArray_get(ba.contents, offset/32);
-    return asm(offset & 0x1f, word) uint { byte };
+    let res = getbuffer8(ba.contents, offset);
+    return res;
```
hkalodner, line 144: This would be better as a direct return since I don't think the compiler optimizes this

stdlib/bytearray.mini
```diff
@@ -129,165 +126,139 @@ public func bytearray_size(ba: ByteArray) -> uint {
...
-        let res = bytearray_get256(ba, offset-ba.sliceOffset);
-        let ret = res / 0x1000000000000000000000000000000000000000000000000;
-        if (bytesToTrim > 0) {
-            ret = ret & ~(asm(8*bytesToTrim, 1) uint { shl } - 1);
-        }
-        return ret;
     }
+    offset = offset + ba.sliceOffset;
+    let res = getbuffer64(ba.contents, offset);
+    return res;
```
hkalodner, line 164: Return directly

stdlib/bytearray.mini
```diff
@@ -129,165 +126,139 @@ public func bytearray_size(ba: ByteArray) -> uint {
...
+        return ba;    // ignore out-of-bounds write
     }
-    offset = offset + ba.sliceOffset;
-    let (newContents, _) = expandingIntArray_op(
-        ba.contents, 
-        offset/32, 
-        struct {
-            f: unsafecast<OpClosureFunc>(basb_closure_func),
-            val: (31-(offset%32), val,),
+    if (offset >= const::MaxByteArraySize) {
```
hkalodner, line 210: This conditional repeats the first and should be corrected

stdlib/bytearray.mini
```diff
@@ -129,165 +126,139 @@ public func bytearray_size(ba: ByteArray) -> uint {
...
-        let (newContents, _) = expandingIntArray_op(
-            ba.contents,
-            lowWordIndex,
-            struct{
-                f: unsafecast<OpClosureFunc>(bas64_single_closure_func),
-                val: (alignment, value,),
-            }
-        );
-        return ba with { contents: newContents };
+    if (offset >= const::MaxByteArraySize - 7) {
```
hkalodner, line 229: This conditional repeats the first and should be corrected

hkalodner (changes requested):

stdlib/bytearray.mini
edfelten: Added a comment explaining this.

stdlib/bytearray.mini
edfelten: Fixed.

stdlib/bytearray.mini
edfelten: Fixed.

hkalodner (approved):
LGTM

PR #221, opened 2020-10-15
Support non eip155 signatures

edfelten:
* Add support for non-EIP155 signatures on transactions
* Change how the `v` component of signatures is compressed in compressed tx headers: EIP155 signatures send `v % 2`, and non-EIP155 signatures send `v` unmodified (which will be 27 or 28).
* Add test of non-EIP155 signatures, based on the signed tx at https://gist.github.com/Agusx1211/de05dabf918d448d315aa018e2572031


arb_os/signedTx.mini
```diff
@@ -27,7 +27,7 @@ import func bytearray_copy(
 import func bytestream_new(ba: ByteArray) -> ByteStream;
 import func bytestream_skipBytes(ba: ByteStream, nbytes: uint) -> option<ByteStream>;
 
-import func rlp_encodeAndHashMessageInfo(tx: SignedTx) -> bytes32;
+import func rlp_encodeAndHashMessageInfoForSignature(tx: SignedTx, eip155v: option<uint>) -> bytes32;
```
hkalodner, line 30: It might be clearer to make this param `eip155ChainID` or just `chainID`

hkalodner (changes requested):

arb_os/signedTx.mini
edfelten: Fixed.

hkalodner (approved):
LGTM

PR #222, opened 2020-10-16
Modules updates

RGates94:
This PR fixes various issues with the module system, and partially ports arb_os to use the system.


arb_os/decompression.mini
```diff
@@ -342,3 +328,7 @@ impure func decompressTx(
     ));
 }
 
+func dumbHack() {
```
edfelten, line 331: Add a comment explaining why this is here.

src/compile/ast.rs
```diff
@@ -12,6 +12,9 @@ use crate::pos::Location;
 use crate::stringtable::StringId;
 use crate::uint256::Uint256;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+pub type TypeTree = HashMap<(Vec<String>, usize), Type>;
```
edfelten, line 17: Add a comment explaining what a `TypeTree` represents.

edfelten (changes requested):
Looks good.  Suggest adding some comments.

arb_os/decompression.mini
RGates94: Added.

src/compile/ast.rs
RGates94: Added.

edfelten (approved):
Looks good.

PR #224, opened 2020-10-16
Fix display of filenames in error messages

RGates94:
This fixes an issue with the module system where the file name in error messages would always be the same as the specifed main file.


edfelten (approved):
Looks good.

PR #225, opened 2020-10-19
Fix bug with unary ops

### LINKED ISSUE #223: Len doesn't work on named types
Doing the Mini `len` operator on an object of named type, where the underlying type is an array, generates a compile-time type error. This construction should be allowed.

This issue was introduced in #222.

Issue arises for code like this:

type FooItem = struct { x: uint, y:uint }
type Foo = []FooItem

func f(foo: Foo) {
    let sz = len(foo);   // compiler reports error here
}

RGates94:
This fixes #223.

### DIFF
--- a/src/compile/typecheck.rs
+++ b/src/compile/typecheck.rs
@@ -1128,7 +1128,7 @@ fn typecheck_expr(
                 return_type,
                 type_tree,
             )?;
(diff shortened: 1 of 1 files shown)


edfelten (approved):
Looks good.

PR #228, opened 2020-10-21
Add fields to tx receipt

edfelten:


arb_os/output.mini
```diff
@@ -217,7 +207,9 @@ public impure func emitLog(
 
     asm((
         0,
-        l1message,
+        l1message with {
+            msgData: unsafecast<ByteArray>(bytearray_marshalFull(l1message.msgData))  // because caller expects marshalled bytes
```
hkalodner, line 211: Is there a good reason why we're not just using a ByteArray here when outputing a log?

arb_os/inbox.mini
```diff
@@ -291,10 +311,7 @@ impure func translateUnsignedTx(request: IncomingRequest) -> option<TxRequestDat
         request = request with {
             requestId: uint(hash(
                 bytes32(request.sender),
-                hash(
-                    bytes32(chainParams_chainId()),
-                    marshalledBytes_hash(request.msgData)
-                )
+                hash(bytes32(chainParams_chainId()), marshalledBytes_hash(bytearray_marshalFull(request.msgData)))
```
hkalodner, line 314: Potentially we should change what hash we use here to be more efficiently computable from a ByteArray

hkalodner (dismissed):
I left a couple comments, for things that we might want to open issues for, but overall LGTM

hkalodner (approved):
LGTM

PR #230, opened 2020-10-22
Fix constructor payment bug

edfelten:
* Fix bug that was preventing payment to constructors
* Fix purity labels in source code
* Provide stipend to cross-contract calls that receive payment
* Disallow storage write if cross-contract call provided less gas than the EVM gas cost of a write


hkalodner (approved):
LGTM. Tested thoroughly against monorepo

PR #232, opened 2020-10-23
Update clap

RGates94:
This updates clap to 3.0.0-beta2.  By updating to this version we gain access to struct based syntax for defining command line arguments.  

While this version is still in beta, there is low risk in adopting it, as another crate, structopt, offers near identical functionality and is stable.  It is preferable to keep using clap rather than structopt, as structopt relies on clap as an external dependency, and this way we need one fewer dependency in the tree.  I also expect that the majority of development going forward will be in the clap crate rather than structopt. 

This branch also updates the benchmark logs, as they were out of date.


edfelten (approved):
Looks good.

PR #233, opened 2020-10-27
Debugging support

edfelten:
When EVM code does a REVERT instruction, emit a debugprint of `(10000, contractAddress, evmPC)` where `contractAddress` is the L2 address of the contract that did the revert, and `evmPC` is the EVM program counter where the revert occurred.


arb_os/arbsys.mini
```diff
@@ -65,12 +67,14 @@ import impure func decompressAddress(stream: ByteStream) -> option<(ByteStream,
...
 
+import impure func pcTableForCode(code: ByteArray) -> option<Stack>;
+
 
 public impure func arbsys_txcall() {
     if let Some(topFrame) = evmCallStack_topFrame() {
         let calldata = evmCallFrame_getCalldata(topFrame);
         if (bytearray_size(calldata) < 4) {
-            evmOp_revert(0, 0);
+            evmOp_revert_knownCodePc(address(100), 0, 0, 0);
```
hkalodner, line 77: It might be better for debugging to give each ArbSys revert it's own artificial PC so that they can be distinguished.

hkalodner (changes requested):

arb_os/arbsys.mini
edfelten: This is done in the next commit.

arb_os/arbsys.mini
```diff
@@ -65,12 +69,14 @@ import impure func decompressAddress(stream: ByteStream) -> option<(ByteStream,
...
```
hkalodner, line 79: It looks like `evmOp_revert_knownCodePc(address(100), 0, 0, 0);` is still used a few times in this file. I'm not sure if that was intentional after the last commit replaced most of them

arb_os/codeSegment.mini
```diff
@@ -516,8 +517,8 @@ public impure func translateEvmCodeSegment(
                     if (opcode == 0xfd) { // REVERT
                         return pushEvmInsnCall(opWrapper2Uint(evmOp_revert), restOfCode, evmJumpTable, pcTable);
```
hkalodner, line 518: Should this also be replaced with `evmOp_revert_knownPc`?

hkalodner (changes requested):

arb_os/codeSegment.mini
edfelten: Fixed.

arb_os/arbsys.mini
edfelten: Fixed.

hkalodner (approved):
LGTM

PR #237, opened 2020-10-28
Improve error message

### LINKED ISSUE #234: Error message should include source file name
The compiler sometimes emits error message like this:
`Warning: expression statement at line: 402 column: 13 returns value of type Uint, which is discarded`

This is nice, except that it should say which source code file the error is in.

RGates94:
This fixes #234 by including the filename in the warning message.


edfelten (approved):
Looks good.

PR #239, opened 2020-10-30
Prevent reuse of addresses

### LINKED ISSUE #238: Deal smoothly with multiple contract deploys at the same address
There is a corner case where two contracts can be deployed at the same address.  (EOA deploys at L1, and deploys with the same nonce at L2, so both deploys happen at the same address. That L1 contract deploys a buddy contract at its same address.)

The desired behavior is that the first contract deployed survives and the second deploy attempt fails.  (Note that either the EOA deploy or the buddy contract deploy can happen first.)

edfelten:
This ensures that if there are multiple attempts to deploy contracts at the same address, only the first one will succeed. Subsequent deploy attempts at the same address will fail, with the tx receipt containing a new failure code.

This scenario can only come up in a rare case where:
(1) an EOA deploys a contract on Ethereum using nonce N
(2) the same EOA deploys a contract on an Arbitrum chain using the same nonce N, and
(3) the Ethereum contract deployed in step 1 deploys a buddy contract on the Arbitrum chain
(Note that the same steps can also happen in other orders.)

We enforce the principle that once a contract is deployed at an address, nothing can dislodge or replace it. The only way the contract can be eliminated is if it self-destructs.

Fixes #238 

This should not require any changes to other modules.

### DIFF
--- a/arb_os/accounts.mini
+++ b/arb_os/accounts.mini
@@ -107,6 +107,10 @@ public func account_isEmpty(acct: Account) -> bool {
     return (acct.nextSeqNum == 0) && (acct.ethBalance == 0) && (acct.contractInfo == None<AccountContractInfo>);
 }
 
+public func account_isContract(acct: Account) -> bool {
+    return (acct.contractInfo != None<AccountContractInfo>);
+}
+
 public func account_checkAndIncrSeqNum(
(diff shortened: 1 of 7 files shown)


hkalodner (approved):
LGTM

PR #240, opened 2020-10-31
Add simplified gas charging mechanism

edfelten:
* Add a new option in the chain initialization message, to submit parameters for gas charging on the chain. If this option is not submitted, gas charging is disabled.
* Charge transactions for gas, based on an adaptive mechanism that adjusts gas price dynamically based on demand and cost.
* Includes some simple testing of the mechanism


edfelten:
This was superseded by #248

PR #241, opened 2020-10-31
Refactor ArbSys

edfelten:
Refactor ArbSys, to move three sets of functionality out into separate contracts:
* `ArbAddressTable`, at address 102, holds the address table and address compression/decompression functions
* `ArbBLS`, at address 103, holds the BLS key registration and lookup functions
* `ArbFunctionTable`, at address 104, holds the functions to support function-table based compression/decompression

This will require changes in the Arbitrum repo, and probably elsewhere, to maintain compatibility with ArbOS.


hkalodner (approved):
LGTM

PR #243, opened 2020-11-12
Fix contract creation precondition

### LINKED ISSUE #242: Creating contract fails if address is already funded
Attempting to create a contract at an address should be allowed, as long as there isn't already a contract at that address.  The current code incorrectly prevent contract creation at an address if the address is already funded, even if it doesn't have a contract yet.   Contract creation should be allowed in that case.

edfelten:
Change the condition for when a contract can be created, to allow contract creation at an address (where the address is chosen according to the standard algorithms) as long as there is not already a contract at the same address.

Previously we would block contract creation if the address was already funded, which is incorrect.

Fixes #242

This should not require any changes in other modules.


arb_os/evmOps.mini
```diff
@@ -272,7 +273,7 @@ public impure func evmOp_extcodehash(addrAsUint: uint) -> bytes32 {
     let addr = address(addrAsUint);     // truncates if necessary
     if let Some(topFrame) = evmCallStack_topFrame() {
         let acct = evmCallFrame_getAccount(topFrame, addr);
-        if (account_isEmpty(acct)) {
+        if (! account_hasContract(acct)) {
```
hkalodner, line 276: I think the old behavior might actually have been correct here, but I'm not sure

edfelten: Yeah, it looks like what is left of this PR (after merging in changes from develop) is incorrect.

edfelten:
Closing this.  The part that is not superseded by other PRs is incorrect.

edfelten:
Closed by mistake.  Reopening and will fix.

arb_os/evmOps.mini
edfelten: Yes, will fix.

hkalodner (approved):
LGTM

PR #245, opened 2020-11-17
Upgradable ArbOS

edfelten:
Add affordances needed to allow runtime upgrading of ArbOS.  
* Add functionality for uploading marshaled AVM code and making it runnable (currently, this is never called). This can be invoked to either allow or disallow privileged AVM instructions. It could also be useful for allowing pluggable AVM-code modules, if we decide to support those.
* Modify EVM->AVM transpiler so that calls to ArbOS core functions in the generated code goes indirectly through a writeable jump table.
* Add ArbSys method to get ArbOS version number.
* Add ArbOwner precompiled contract, at address 107, giving the chain's owner the ability to transfer ownership, or to upload upgrade code and jump to it.  Upload of code can happen in multiple chunks, to allow larger upgrades to be uploaded.


hkalodner (approved):
LGTM

PR #246, opened 2020-11-20
Add field to block summary log giving number of previous block

edfelten:
Add a field to the end of the block summary receipt, giving the block number of the previous block that had a block summary receipt.  If there was no such previous block, the field will be MaxUint256.

This might or might not require changes to aggregator and block explorer software, depending on how they parse the block summary receipts.


arb_os/output.mini
```diff
@@ -85,11 +86,14 @@ public impure func outputStats_endOfBlock(blockNum: uint, timestamp: uint) {
         updatedStats = updatedStats with {
             numLogs: updatedStats.numLogs + 1  // +1 for soon-to-be-issued block summary log
         };
+
+        emitBlockSummaryLog(blockNum, timestamp, globalBlockReceiptData.lastBlockNum);
+
         globalBlockReceiptData = globalBlockReceiptData with {
             statsAllTime: updatedStats
+        } with {
+            lastBlockNum: blockNum
```
hkalodner, line 95: This reorders the call to `emitBlockSummaryLog` before the update to `globalBlockReceiptData`, but `emitBlockSummaryLog` depends on the value of that global so this effects the block logs. I think the original order should be restored

hkalodner (changes requested):

arb_os/output.mini
edfelten: Yes, that's a bug. Fixed it by splitting the update to globalBlockReceipt data into two pieces, one before emitting the block summary and one after.

hkalodner (approved):
LGTM

PR #248, opened 2020-11-23
Add mechanism to charge clients for gas and aggregator services

edfelten:
* Add a new option in the chain initialization message, to submit parameters for gas charging on the chain. If this option is not submitted, gas charging is disabled.
* Charge transactions for gas, based on an adaptive mechanism that adjusts gas price dynamically based on demand and cost.
* Add ability of aggregators to set fees, and clients to agree to pay those fees. Collect fees as part of the ArbGas charging mechanism.

Includes some simple testing of the mechanisms.

This will require changes to aggregators and clients.


edfelten:
Closing this because we chose a different approach to denominating gas.

PR #249, opened 2020-11-23
Vm hash instruction

edfelten:
Add a new AVM instruction, "vmhash", opcode 0x7c, which pushes the hash of the VM (as it was before the instruction started executing) onto the stack.  Gas cost is zero.

Support this in the Mini compiler, via the `asm` construct.

### DIFF
--- a/builtin/globaltest.mini
+++ b/builtin/globaltest.mini
@@ -125,6 +125,11 @@ impure func tests() -> uint {
         return 22;
     }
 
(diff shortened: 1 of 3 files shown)


edfelten:
Turning this back into a draft, because it's not currently needed and it might have more significant implications elsewhere in the stack.

edfelten:
Closing this because it's stale and we don't think we'll need this functionality.

PR #250, opened 2020-11-23
Inlining groundwork

RGates94:
This PR adds a partial implementation of inlining, with the option to inline code disabled.


edfelten (approved):
Looks good.

PR #252, opened 2020-11-24
Fix incorrect handling of INVALID EVM instruction

edfelten:
This fixes a regression in #251 .  The change in that PR caused the EVM transpiler to generate incorrect code for the EVM INVALID instruction.


hkalodner (approved):
LGTM

PR #253, opened 2020-11-24
Rescale ArbGas by a factor of 1000

edfelten:
Rescale ArbGas by a factor of 1000, so what used to be 100,000 ArbGas is now 100 ArbGas.  

Basically this treats the AVM's ArbGasRemaining register as if it measures milli-ArbGas, with 1000 units of the register corresponding to 1 ArbGas.  ArbOS's data structures keep track of the rescaled gas--only the AVM register uses the original scaling.

One consequence of this is that transactions will use much less ArbGas, compared to the amount of L1 gas they would require. So dapps that give low amounts of gas to limit what callees can do probably won't enforce those limits effectively. (An exception is that a zero-gas transaction that makes an ETH transfer will be prevented from modifying storage, even though it will receive an automatic gas stipend.)

This will require changes to the AVM specification, to reduce the gas usage of all operations by a factor of 1000 (and to say that the gas usage of an activity is rounded up to an integer).


edfelten:
Closing this because we chose a different approach to denominating gas.

PR #255, opened 2020-12-02
Refactor ast

RGates94:
This PR refactors the way that the data in the AST is structured, splitting each node into a main struct containing DebugInfo, and a kind enum without debug information.


edfelten (approved):
Looks good.

PR #256, opened 2020-12-04
Bytearray size limit

edfelten:
Cap the size of bytearrays at 2**64 bytes.  This eliminates weird behaviors associated with overflow of sizes and offsets.
* unmarshalling a byte stack that claims to be larger returns an error
* bytearray_set* ignores writes that would touch offsets >= 2**64 
* bytearray_copy shortens a copy if necessary so that the result fits within the limit

This PR also fixes some minor issues with duplicate error codes and unneeded use statements.


hkalodner (approved):
LGTM

PR #257, opened 2020-12-04
Add specialized type to improve typechecking

edfelten:
Add specific types, rather than using the generic queue types, in inbox.mini. This allows the compiler to do better type-checking. 

If/when Mini supports generics, we can eliminate this. Until then, this is a reasonable compromise, especially given that inlining will soon reduce the cost of this.


hkalodner (approved):
LGTM

PR #259, opened 2020-12-09
Debugging improvements

RGates94:
This adds the ability to mark statements as a breakpoint from within the mini code.  This will stop execution in the debugger when you reach an instruction with this attribute.


edfelten (approved):
Looks good.  I'm looking forward to using this.

PR #260, opened 2020-12-10
Deposit with tx

edfelten:
This creates a new L1 message type, representing an ETH deposit that also executes a transaction at L2.  It's basically like a payable transaction at L2, where the funds are provided by the sender at L1 (rather than at L2 as we normally do).

This is done via a new L1 message type 7, "deposit with transaction".  Its type-specific data must be an L2 message of subtype 0 (unsigned message from EOA) or 1 (unsigned message from contract). The EthBridge is assumed to have received and locked the funds.

The semantics are of an unconditional deposit into the sender's L2 account, followed by executing the transaction at L2. If the L2 transaction reverts, the deposit has still happened, so the funds will be left in the caller's L2 account.  (The caller can withdraw those funds if it wants, using a plain ETH withdrawal or a liquidity provider.)

The transaction executes normally and generates a normal receipt.

The old ETH deposit message is still supported for backward compatibility, but the old message type will presumably be deprecated.


hkalodner (approved):
LGTM

PR #262, opened 2020-12-10
Change file_name_chart from HashMap to BTreeMap

RGates94:
This fixes the lack of determinism in the compiler output.  Previous versions would have different orders for files in the file name chart, but were otherwise deterministic.


edfelten (approved):
Looks good.

PR #263, opened 2020-12-13
Refactor and improve EVM-AVM codegen

edfelten:
* Refactor AVM->EVM translator to eliminate recursion. The new structure makes a forward parsing pass to build a stack of parsed EVM instructions, then a backward codegen pass using that stack.
* Avoid generating code for EVM instructions that are unreachable (after an instruction that diverts control flow, without an intervening JUMPDEST).
* Change stack implementation in library to store 8 items per cell rather than one.
* Use eager hashing of tuples in the Rust emulator, to avoid deep recursion in the emulator when hashing deep tuple structures.  (Rust doesn't like deep recursion.)

This passes the big EVM compatibility test suite.


src/mavm.rs
```diff
@@ -277,33 +278,41 @@ impl fmt::Display for CodePt {
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub enum Value {
     Int(Uint256),
-    Tuple(Rc<Vec<Value>>),
+    Tuple(Rc<Vec<Value>>, Uint256),
```
hkalodner, line 281: This change causes the JSON serialization format to be changed, which I don't think we want. This will probably require changing the serialization format of the `Tuple` type

hkalodner (changes requested):

arb_os/codeSegment.mini
```diff
@@ -154,261 +155,316 @@ public impure func pcTableForCode(code: ByteArray) -> option<Stack> {
...
+                                } elseif (opcode == 0x0b) { // SIGNEXTEND
+                                    code = pushInstruction(0x0b, code);
+                                } else {
+                                    invalidInstruction = true;
+                                }
+                            }
+                        } else {
+                            if ( (opcode >= 0x10) && (opcode <= 0x1a) ) {
+                                code = pushInstruction(opcode, code);
+                            } elseif (opcode == 0x1b) { // SHL
```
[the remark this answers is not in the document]
edfelten: Good catch. Fixed this.

src/mavm.rs
edfelten: Switched back to the old structure. This is viable because of the change to the ArbOS stack library.

hkalodner (approved):
LGTM

PR #264, opened 2020-12-16
Use named constants in the ArbOS source code

edfelten:
* Add a list of pre-defined constants that are available in every Mini program
* Constants can also be defined in a Mini source code file, after the `use` statements, using syntax like `const my constant = 42;`
* At present, constants must be unsigned integers, but they could be extended in principle to support signed int, boolean, and None, and empty-tuple
* Allow use of a named constant anywhere an unsigned integer literal is allowed, using the syntax `const::name`
* Use constants throughout the ArbOS source code, for clarity

Some compromises were made to reduce implementation complexity. For example, the predefined constants are in the compiler's source code, rather than defined in Mini syntax and parsed by the compiler. Also, constants defined in one file can't be referenced in another file. And we require constants to be unsigned integers so they can easily be used in places where unsigned integers are required, such as declaring the sizes of fixed-size arrays.


arb_os/arbsys.mini
```diff
@@ -437,52 +437,52 @@ impure func arbAddressTable_addressExists(topFrame: EvmCallFrame, calldata: Byte
...
+        evmOp_revert_knownCodePc(address(const::Address_ArbAddressTable), 160, 0, 0);
     }
     let offset = bytearray_get256(calldata, 36);
     let bytesLen = bytearray_get256(calldata, 68);
     if (offset >= bytesLen) {
-        evmOp_revert_knownCodePc(address(102), 161, 0, 0);  // buffer is too short
+        evmOp_revert_knownCodePc(address(const::Address_ArbAddressTable), 161, 0, 0);  // buffer is too short
     }
-    let stream = bytestream_new(bytearray_extract(calldata, 100+offset, bytesLen-offset));
+    let stream = bytestream_new(bytearray_extract(calldata, const::Address_ArbSys+offset, bytesLen-offset));
```
hkalodner, line 485: This doesn't seem like a correct use of this constant

arb_os/arbsys.mini
```diff
@@ -499,18 +499,18 @@ impure func arbAddressTable_decompress(topFrame: EvmCallFrame, calldata: ByteArr
...
-        evmOp_revert_knownCodePc(address(102), 163, 0, 0);
+        evmOp_revert_knownCodePc(address(const::Address_ArbAddressTable), 163, 0, 0);
     }
 }
 
 impure func arbAddressTable_compress(topFrame: EvmCallFrame, calldata: ByteArray) { // (address addr) -> bytes
 // compress addr, based on the current address table contents, and return the resulting buffer
     if (bytearray_size(calldata) != 36) {
-        evmOp_revert_knownCodePc(address(102), 100, 0, 0);
+        evmOp_revert_knownCodePc(address(const::Address_ArbAddressTable), const::Address_ArbSys, 0, 0);
```
hkalodner, line 513: This seems like another incorrect use of `const::Address_ArbSys`

arb_os/evmCallStack.mini
```diff
@@ -216,16 +216,33 @@ public impure func initEvmCallStack(
...
+                        const::TxResultCode_insufficientBalance,
+                        None<ByteArray>,
+                        None<EvmLogs>,
+                        Some(gasUsage)
+                    );
                 } else {
                     // unknown error type
-                    emitTxReceipt(request.incomingRequest, 255, None<ByteArray>, None<EvmLogs>, None<GasUsage>);
+                    emitTxReceipt(
+                        request.incomingRequest, const::TxResultCode_unknownFailure,
```
hkalodner, line 229: The formatting here should probably split this line

hkalodner (changes requested):

arb_os/arbsys.mini
edfelten: Fixed

arb_os/arbsys.mini
edfelten: Fixed.

arb_os/evmCallStack.mini
edfelten: Fixed.

hkalodner (approved):
LGTM

PR #265, opened 2020-12-16
Ripemd-160 hash function and precompile

edfelten:
* Implement `ripemd160f` instruction (RIPEMD-160 compression function) instruction in Rust AVM and Mini compiler
* Add RIPEMD-160 functionality to the ArbOS library
* Support RIPEMD-160 EVM precompile at address 0x03
* Testing for library and precompile


Cargo.toml
```diff
@@ -34,3 +34,4 @@ ethers-signers = "0.1.3"
 keccak="0.1.0"
 bn-plus = "0.4.4"
 rust-crypto="^0.2"
+ripemd160="0.9.1"
```
hkalodner, line 37: I'm guessing this is the crate that `ripemd160port.rs` came from, but I don't see any actual usages of this crate in the code. If there aren't any, this should probably be removed as a dependency from the Cargo file

hkalodner (changes requested):

src/run/ripemd160port.rs
```diff
@@ -0,0 +1,295 @@
+use core::convert::TryInto;
+
+pub const DIGEST_BUF_LEN: usize = 5;
+pub const WORK_BUF_LEN: usize = 16;
+
+//type Block = GenericArray<u8, U64>;
+type Block = Vec<u8>;
+
+macro_rules! round(
```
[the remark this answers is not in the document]
edfelten: Added a pointer to the source.

Cargo.toml
edfelten: Removed

hkalodner (approved):
LGTM

PR #266, opened 2020-12-17
Modexp precompile and bignum library

edfelten:
* Add bignum functionality to the ArbOS library. This supports standard operations on arbitrary-size unsigned integers.
* Implement EVM precompile 0x05 (modular exponentiation), using the bignum library.
* Add tests for the bignum library and precompile 0x05.


hkalodner (approved):
I've reviewed everything other than biguint.mini and done some blackbox testing on the modexp implementation, but have not done a full review of the bignum library implementation. The modexp precompile implementation appears to have a quite high gas cost and should be optimized in the future

I think we can merge as is, and add further testing later

PR #267, opened 2020-12-17
Fix bug in linker

RGates94:
This fixes a bug in the linker that was causing incorrect behavior at runtime.


edfelten (approved):
Looks good.  Thanks for tracking this down!

PR #268, opened 2020-12-18
Fix incorrect behavior of EVM ORIGIN instruction

edfelten:
Fix incorrect behavior of the EVM ORIGIN instruction.


hkalodner (approved):
LGTM

PR #269, opened 2020-12-18
Use recursive types in the mini code and remove unused files

RGates94:
This changes fields of type `any` to recursive types to improve type safety.


edfelten:
Is it possible to remove more unsafecasts from the files that were changed in this PR?  It looks to me like there are some that could be eliminated.

RGates94:
> Is it possible to remove more unsafecasts from the files that were changed in this PR? It looks to me like there are some that could be eliminated.

Yes, I didn't notice these when I was first writing this PR, but looking back at it there are definitely some casts that can be removed.  I'll make these changes by Monday.

edfelten (approved):
Looks good.

PR #270, opened 2020-12-20
Fix EC precompiles

hkalodner:
Currently the ECADD and ECMUL precompiles are implemented to revert when the input is not 128 bytes. Instead the function should support any input size, filling with zeros for reads beyond the input


edfelten (approved):
Looks good.  Thanks for catching this.

PR #271, opened 2020-12-20
Debugprint more info on error or REVERT

edfelten:
Print out more useful information on errors and EVM reverts.

On revert, debugprint (1000, codeAddress, EVM PC, txid, currentAddresses, parentAddresses).
On out-of-gas, debugprint (665, txid, currentAddresses, parentAddresses).
On other error, debugprint (666, txid, currentAddresses, parentAddresses).
On cascading errors, will sometimes debugprint just 666.

txid is the transaction ID of the currently running transaction, or zero if there isn't one.

currentAddresses is (address of account we're running in, address whose code we're running), or (0, 0) if we're not in a contract.

parentAddresses is the same pair of addresses, but for the parent EVM call frame, or (0, 0) if there isn't one.


arb_os/errorHandler.mini
```diff
@@ -34,22 +38,32 @@ public impure func errorHandler() {
...
+        inErrorHandler = true;
+        if (asm() uint { getgas } > (~0 - 500)) {
+            // application ran out of ArbGas
+            let (txid, addrs, parentAddrs) = evmCallStack_stackInfo();
+            asm((665, txid, addrs, parentAddrs),) { debugprint };
+        } else {
+            // some other error
+            let (txid, addrs, parentAddrs) = evmCallStack_stackInfo();
+            asm((666, txid, addrs, parentAddrs),) { debugprint };
+        }
```
hkalodner, line 54: Probably better to lift the `let (txid, addrs, parentAddrs) = evmCallStack_stackInfo();` out of the if-else to reduce code duplication unless I'm missing something

hkalodner (changes requested):

arb_os/errorHandler.mini
edfelten: This is deliberate. The condition on the if statement depends on the amount of ArbGas used. I wanted to call evmCallStack_stackInfo after checking the gas condition, rather than before. That way we don't have to worry about whether the ...stackInfo call changes the result of the gas comparison.

hkalodner: Ahhh. That makes a ton of sense 😁 

It'd probably be good to add a comment to the code explaining that reasoning so that somebody in the future doesn't accidentally break it

edfelten: Done.

hkalodner (approved):
LGTM

PR #273, opened 2020-12-22
Merkleize sends

edfelten:
Rather then emitting every send individually, this PR instead emits a log (of a new log type 2) containing what would be sent.  Periodically, a true send is done, containing a Merkle root of the items that would have been sent since the last Merkle-root send.


arb_os/output.mini
```diff
@@ -227,6 +238,8 @@ impure func emitBlockSummaryLog(blockNum: uint, timestamp: uint, prevBlockNum: u
         globalBlockReceiptData.statsAllTime,
         prevBlockNum
     ),) { log };
+    asm(merkleTreeBuilder_rootHash(merkleOfSends),) { send };
```
hkalodner, line 241: As discussed offline, we may want to reduce the frequency of sending Merkle roots depending on the cost of processing a single Merkle root

hkalodner (approved):

PR #274, opened 2020-12-22
Support L2 code sending calls to L1

edfelten:
Add an ArbSys method that sends a call to the L1. 

```function sendTxToL1(address destAddr, bytes calldataForL1) external payable;```

This will cause the L2 to do a send (2, callerAddressAtL2, data), where data is:
* destination address at L1 (32 bytes)
* current block number when L2 made the call (32 bytes)
* current timestamp when L2 made the call (32 bytes)
* callvalue (32 bytes)
* calldata (all remaining bytes)

If callvalue is non-zero, those funds are being withdrawn from L2, and should be unlocked at L1 and used as callvalue of the L1 call.


hkalodner (approved):
LGTM

PR #275, opened 2020-12-23
Handle unknown default values rather than panicking

edfelten:
Change `Type::default_value` to return a pair `(defaultValue, isTypeSafe)` rather than panicking when there isn't a known default value. This allows the code generator to (correctly, according to the Mini spec) create new arrays of any type, with the array slots being uninitialized if the type doesn't have a default value.

### DIFF
--- a/src/compile/codegen.rs
+++ b/src/compile/codegen.rs
@@ -1633,7 +1633,7 @@ fn mavm_codegen_expr<'a>(
                 vec![Type::Uint, Type::Any],
                 Box::new(array_type.clone()),
             );
-            let default_val = base_type.default_value().unwrap_or(Value::none());
+            let (default_val, _is_safe) = base_type.default_value();
             let the_expr = TypeCheckedExpr {
                 kind: TypeCheckedExprKind::FunctionCall(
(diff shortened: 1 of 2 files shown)


src/compile/codegen.rs
```diff
@@ -1633,7 +1633,7 @@ fn mavm_codegen_expr<'a>(
                 vec![Type::Uint, Type::Any],
                 Box::new(array_type.clone()),
             );
-            let default_val = base_type.default_value();
+            let default_val = base_type.default_value().unwrap_or(Value::none());
```
RGates94, line 1636: As currently written it looks like we always change a `None` return value to `Value::none()`, if we don't expect to change this soon it would be simpler to simply return `Value::none()` directly, and change the signature of `default_value` from `Option<Value>` back to `Value`

RGates94 (changes requested):

src/compile/codegen.rs
edfelten: Revised based on out-of-band conversation.

RGates94 (approved):
Looks good to me.

PR #278, opened 2020-12-24
Comply with EIP 150

edfelten:
Comply with EIP 150, which requires that a caller can't give a caller more than 63/64 of the caller's gas.


hkalodner (approved):
LGTM

PR #279, opened 2020-12-25
Fix invalid code elimination (selfbalance opcode)

hkalodner:


edfelten (approved):
Looks good.  This opcode isn't listed in some sources, but it does exist in the current version of Ethereum.

PR #280, opened 2020-12-26
Fix staticcall behavior

edfelten:
Change the behavior of STATICCALL EVM instruction to be consistent with EIP-214.  Track a static flag in each EVM call frame, and revert on mutating operations if the static flag is set.  This required some minor refactoring.


hkalodner (approved):
LGTM

PR #282, opened 2020-12-27
Eliminate imports

edfelten:
* Eliminate duplicate type declarations
* Replace obsolete `import` statements with equivalent `use` statements
* Remove obsolete recipes from Makefile


edfelten:
Merging this without review, because the changes are minimal and don't change semantics.

PR #283, opened 2020-12-28
Partially fix inlining

RGates94:
This resolves a number of bugs with the existing inlining implementation.  There are still issues, and cross file inlining has been temporarily disabled, however arbos now compiles with inlining turned on everywhere, and all tests that do not rely on arbos.mexe or biguinttest.mexe pass with inlining enabled everywhere.


edfelten (approved):
Looks good, as far as I can see.  We should have a deeper dive discussion about inlining and how to optimize it.

PR #285, opened 2020-12-28
Pluggable modules for ArbOS

edfelten:
Add a facility to have pluggable modules in ArbOS.  A module is a function with signature `impure func(args: any, state: any, neverBeenCalledBefore: bool) -> (returnVal: any, updatedState: any)`.   

A module manager stores functions under integer indexes, stores a state for each function, and allows calls by index.

A new module can be plugged in at an existing index, with the option to either keep the old module's state or set the state to `()`. 

Functionality in `ArbOwner` which previously supported code uploads to upgrade ArbOS, now supports code uploads for either upgrading all of ArbOS, or uploading and installing a pluggable module.


hkalodner (changes requested):
This looks good to me other than the small suggestion. The main thing this PR is missing is an applied use of the pluggable modules system, but that can be added separately

hkalodner (approved):
LGTM

PR #288, opened 2020-12-28
Automatically refund tokens if ERC20/721 deposit fails

edfelten:
For every ERC20/721 deposit, this prepares a withdraw message to send those tokens back to their L1 sender. That withdraw message is sent if the deposit fails.


hkalodner (approved):
LGTM

PR #289, opened 2020-12-28
Pluggable rollup tracker

edfelten:
Make the rollup tracker a pluggable module.  This refactors the rollup tracker, and then accesses it through the pluggable module API.  This will allow chain owners to upload new versions of the rollup tracker over time.


hkalodner (approved):
I like this a lot

PR #290, opened 2020-12-29
Let pluggables call pluggables, and let owner bind a pluggable to an address

edfelten:
* Change the API for calling pluggables, to pass in a function pointer that can be used to call other pluggables.
* Allow the chain owner to bind a pluggable to an address, so that a contract call to that address will go to the pluggable. This enables real-time uploading of trusted contracts written in Mini, and makes those contracts hot-swappable.
* As a proof of concept, this installs precompile 0x05 (modexp) as a pluggable


hkalodner (commented):
This looks great so far

arb_os/precompiles.mini
```diff
@@ -256,6 +272,45 @@ public impure func precompile_0x05() {  // modexp
     evmOp_revert_knownPc(2, 0, 0);
 }
 
+func precompile_0x05_pluggable(
+    args: (EvmCalFrame, ByteArray, any),
```
hkalodner, line 276: Looks like this is a typo? `EvmCalFrame`. Shouldn't this have failed compilation?

arb_os/precompiles.mini
```diff
@@ -256,6 +272,45 @@ public impure func precompile_0x05() {  // modexp
...
+    state: any,
+    firstTime: bool
+) -> (any, any) {  // return ((success: bool, returnData: ByteArray), any)
+    let (topFrame, calldata, _) = args;
```
hkalodner, line 280: Looks like `topFrame` isn't used

arb_os/precompiles.mini
```diff
@@ -256,6 +272,45 @@ public impure func precompile_0x05() {  // modexp
...
+
+    let base = bytearray_extract(calldata, 96, baseLen);
+    let exp = bytearray_extract(calldata, 96+baseLen, expLen);
+    let mod = bytearray_extract(calldata, 96+baseLen+expLen, modLen);
+
+    if let Some(result) = biguint_modexp(
+        biguint_fromBytes(base),
+        biguint_fromBytes(exp),
+        biguint_fromBytes(mod),
+    ) {
```
hkalodner, line 294: Is the thought that all of the function calls would be compiled as part of the code segment for this pluggable, or would it use the existing implementation?

hkalodner (changes requested):

arb_os/precompiles.mini
edfelten: Yeah, that's odd.  I raised the compiler issue with Robert, and fixed the typo.

arb_os/precompiles.mini
edfelten: Replaced with _

arb_os/precompiles.mini
edfelten: At present a newly uploaded pluggable would have to have its own bytearray library compiled in.  In the future it would be nice to have linkage information available dynamically to pluggables via a lookup table.  But that's too heavy a lift to do at the moment.

hkalodner (approved):
LGTM

PR #291, opened 2020-12-29
Fix underfunded call bug

edfelten:
Fix a bug that was causing a panic when a cross-contract payable call reverted because the caller had insufficient funds.

Also, refactor the ArbOS test contracts.


hkalodner (approved):
LGTM

PR #292, opened 2020-12-29
Remove old imports

RGates94:
This removes the last imports using the old system, and disables the ability to parse imports of this type.


edfelten (approved):
Looks good.

PR #293, opened 2020-12-29
Constructor value

hkalodner:
- Make external constructor calls properly setup call value based on the passed in value
- Fix bug in test surfaced by the ArbOS fix


edfelten (approved):
Looks good.

PR #296, opened 2020-12-29
Support BLS-signed message batches

edfelten:


src/evm/mod.rs
```diff
@@ -359,7 +365,7 @@ pub fn evm_test_arbsys_direct(log_to: Option<&Path>, debug: bool) -> Result<(),
     assert_eq!(my_addr.clone(), my_addr_decompressed);
     assert_eq!(offset, Uint256::from_usize(my_addr_compressed.len()));
 
-    assert_eq!(Uint256::from_u64(2), arb_address_table.size(&mut machine)?);
+    assert_eq!(Uint256::from_u64(3), arb_address_table.size(&mut machine)?);
```
hkalodner, line 368: I don't see any clear reason why this has changed?

arb_os/decompression.mini
```diff
@@ -223,6 +235,74 @@ public impure func decompressAndVerifyEcdsaSignedTx(
...
+
+        let aggregator = originalRequest.sender;
+        if let Some(aggInfo) = originalRequest.aggregator {
+            aggregator = aggInfo.aggregator;
+        }
+
+        let (_, newTx) = decompressTx(bytestream_new(txData), Some(signerAddr), None<FunctionTable>)?;
+        newTx = newTx with {
+            incomingRequest: originalRequest with {
+                requestId: uint(hash(bytes32(originalRequest.requestId), bytes32(i)))
```
hkalodner, line 286: BLS transactions should have their request ID be calculable before they enter into the inbox like ECDSA transactions

hkalodner (changes requested):

stdlib/bls.mini
```diff
@@ -40,24 +123,943 @@ public func bls_marshalPublicKey(key: BLSKey, ba: ByteArray, offset: uint) -> By
...
     numMsgs:   uint,
-    messages:  []ByteArray,
+    messages:  []BLSHashedMessage,
     keys:      []BLSKey,
     signature: BLSSignature
-) -> bool {
-    return true;
-}
\ No newline at end of file
+) -> option<()> {
```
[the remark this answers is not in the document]
edfelten: Removed it.

src/evm/mod.rs
edfelten: There was a change in decompression.mini, line 155, which causes the decompressor to add addresses to the address table more aggressively.  In this test case, one more address gets added, compared to before.

arb_os/decompression.mini
edfelten: Changed this to have the request ID be the keccak256 of the data that the user signed.

hkalodner (approved):
LGTM

PR #297, opened 2020-12-30
Fix function arg bug

RGates94:
Fixes a bug where function arguments of undefined types might not be caught.


edfelten (approved):
Looks good.

PR #299, opened 2020-12-30
change bls batch size + formatting

sgoldfed:
- Changing bls batch size from 30 to 29
- minor formatting


edfelten (approved):
Good catch.

PR #301, opened 2020-12-31
Fix sequence number increment on constructor errors

edfelten:
Fix the handling of the sender's sequence number on a deploy transaction. Previously we incremented the sequence number in some error cases where it should not be incremented.

Also, return a more accurate error code when a deploy's caller has insufficient funds for calldata.


hkalodner (approved):
LGTM

PR #304, opened 2021-01-04
Type display

RGates94:
This adds a method `display` to `Type` that returns a string representing how that type would appear in mini code.  There are a few formatting limitations:

1. A trailing comma followed by a space always appears where it is possible.
2. Struct fields are always indented by four spaces, including sub-structs.
3. Named types are identified by path and ID.

For example the type `struct { x: (uint, option<[4]int>)}` will return:
```
struct {
    x: (uint, option<[4]int>, ),
}
```

This method has been used to include type information with various error messages, and replaces debug formatting for others.  This method is not implemented as a `Display` impl, as to resolve named types we will need to add an additional argument.

### DIFF
--- a/src/compile/typecheck.rs
+++ b/src/compile/typecheck.rs
@@ -1656,19 +1656,11 @@ fn typecheck_expr(
                     type_tree,
                     scopes,
                 )?;
-                if tc_sub1.get_type() != Type::Bool {
+                if (tc_sub1.get_type(), tc_sub2.get_type()) != (Type::Bool, Type::Bool) {
                     return Err(new_type_error(
                         format!(
-                            "operands to logical and must be boolean, got {}",
-                            tc_sub1.get_type().display()
-                        ),
-                        loc,
-                    ));
-                }
-                if tc_sub2.get_type() != Type::Bool {
-                    return Err(new_type_error(
-                        format!(
-                            "operands to logical and must be boolean, got {}",
+                            "operands to logical and must be boolean, got \"{}\" and \"{}\"",
+                            tc_sub1.get_type().display(),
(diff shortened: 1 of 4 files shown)


src/compile/typecheck.rs
```diff
@@ -1613,13 +1624,19 @@ fn typecheck_expr(
                 )?;
                 if tc_sub1.get_type() != Type::Bool {
                     return Err(new_type_error(
-                        "operands to logical or must be boolean".to_string(),
+                        format!(
```
edfelten, line 1627: It seems better to print both operand types here, so the user doesn't have to guess which one has the wrong type.

src/compile/typecheck.rs
```diff
@@ -1649,13 +1666,19 @@ fn typecheck_expr(
                 )?;
                 if tc_sub1.get_type() != Type::Bool {
                     return Err(new_type_error(
-                        "operands to logical and must be boolean".to_string(),
+                        format!(
```
edfelten, line 1669: Same as above: better to print both operand types.

src/compile/typecheck.rs
```diff
@@ -1921,7 +1951,10 @@ fn typecheck_expr(
                             ))
                         } else {
                             Err(new_type_error(
-                                "fixedarray index must be Uint".to_string(),
+                                format!(
+                                    "fixedarray index must be Uint, found \"{}\"",
```
edfelten, line 1955: uint shouldn't be capitalized

edfelten (changes requested):
A few small changes requested. Overall, looks good.

edfelten (approved):
Looks good.

PR #306, opened 2021-01-05
Fix develop

RGates94:
Ed's integration fix is necessary to have correct execution in both next and develop, so it should be available in both.


edfelten (approved):
Looks good.

PR #308, opened 2021-01-06
Xform speedup

RGates94:
This changes the signatures of `read_code` and `write_code` to prevent copying of their inputs, and adds a line printing the total time spent by the compiler.  This improves overall compile times significantly,  on my local machine, the time to compile ArbOs decreased from 111.805 seconds to 1.119 seconds.


edfelten (approved):
Looks good.  Very happy to have a much faster compiler.

PR #309, opened 2021-01-08
Send every n blocks

edfelten:
* Add a new "outbox" functionality, which handles sends.
* The outbox accumulates sends and emits them in batches. Each batch has a batch number, which starts at zero and increments for each subsequent batch.
* The interval between when each batch is sent is a new chain-specific parameter, which is initially set to a default value and can be modified via an ArbOwner call.
* A send is now a byte string, containing a 32-byte send type followed by type-specific data.
* When a send is put into the outbox, an AVM log is emitted as the 5-tuple (2, batchNumber, indexInBatch, size, buffer)
* When it is time to emit a batch, a log is emitted containing the full batch contents, as the 4-tuple (3, batchNumber, numSendsInBatch, batchContents) where batchContents is a binary tree containing the sends, with each leaf being a 3-tuple (size, buffer, () ).  The binary tree has the same shape as the Merkle tree that summarizes the batch.
* When a batch is emitted, ArbOS does an AVM send, sending a (size, buffer) pair, where the size is 96 and the buffer contains: batchNumber (32 bytes), numSendsInBatch (32 bytes), Merkle root (32 bytes).  The Merkle root computes the hash of each leaf as hash(hash(bytes)), where bytes is the byte-string represented by the (size, buffer) pair in that leaf. The leaf hashes are combined into a Merkle root by the same algorithm as the Solidity Merkle tree library used in the EthBridge.


arb_os/chainParameters.mini
```diff
@@ -52,10 +53,11 @@ public impure func chainParams_gotParamsMessage(sender: address, data: ByteArray
             baseStake: bytearray_get256(data, 3*32),
             stakingToken: address(bytearray_get256(data, 4*32)),
             owner: address(bytearray_get256(data, 5*32)),
+            blocksPerSend: bytearray_get256(data, 6*32),
```
hkalodner, line 56: Is there some way we could set this to a reasonable secure default and then have an extra option to customize it?

hkalodner (changes requested):

arb_os/chainParameters.mini
edfelten: Fixed this.

stdlib/merkletree.mini
```diff
@@ -45,7 +46,7 @@ func merkleize2(data: []any, offset: uint, num: uint, chunk: uint) -> bytes32 {
         return hash(
             merkleize2(data, offset, num1, chunk/2),
             merkleize2(data, offset+chunk, num2, chunk/2),
-        );
+        ) ^ bytes32(1);        // flip low-order bit in hash of internal node, to avoid collision with leaf hash
```
hkalodner, line 49: Rather than flipping the internal nodes, can we flip the leaves instead? That way we wont need to change our merkle proof verification code that's already written in solidity and used in the challenges

hkalodner (changes requested):
LGTM

hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #310, opened 2021-01-11
Remove imported type uses

RGates94:
This removes logic that is no longer necessary with the new module system


edfelten (approved):
Looks good.

PR #311, opened 2021-01-11
Change send opcode to take [size, buffer] arguments

edfelten:
* Change the `send` opcode to take two stack arguments [size, buffer] and output a byte-string (like a Solidity `bytes` object).
* Update Rust client and tests to account for this change
* Change ArbOS to account for this change.  Instead of sending a single `bytes32` object, this now sends a byte-string of size 32.

The Rust VM currently does not check that the `size` argument is within bounds.  (It will always be within bounds for ArbOS, because all ArbOS sends will be of size 32.)


edfelten:
This will be merged into another PR.

PR #312, opened 2021-01-11
L2 message uses buffer

edfelten:
* Change the format for incoming messages to a chain, to the type-specific data field, which was a "bytestack" marshaled structure, with two fields: a size and a buffer.
* Update ArbOS and the Rust test harness to account for the change.


edfelten:
This will be merged into another PR.

PR #313, opened 2021-01-12
Remove symtable

RGates94:
Removes symtable.rs by changing uses of this struct into HashMaps.  While there is an added performance cost when creating a new scope, the cost of looking up local or function is reduced.


edfelten (approved):
Looks good.

PR #314, opened 2021-01-13
Function Contravariance

RGates94:
For soundness it is necessary for function arguments and return types to be contravariant, this PR makes subtyping behave in this way.


src/compile/ast.rs
```diff
@@ -194,8 +194,8 @@ impl Type {
             Type::Func(is_impure, args, ret) => {
                 if let Type::Func(is_impure2, args2, ret2) = rhs {
                     (*is_impure || !is_impure2)
-                        && arg_vectors_assignable(args, args2, type_tree, seen.clone())
-                        && (ret2.assignable(ret, type_tree, seen)) // note: rets in reverse order
+                        && arg_vectors_assignable(args2, args, type_tree, seen.clone())
+                        && (ret.assignable(ret2, type_tree, seen)) // note: rets in reverse order
```
edfelten, line 198: Should the comment "note: rets in reverse order" be revised?

edfelten (commented):
One question to resolve before approving this.

src/compile/ast.rs
RGates94: Yes, I'll edit this.

edfelten (approved):
Looks good.

PR #316, opened 2021-01-13
Eliminate use of byte stacks in ArbOS input/output

edfelten:
* Change ArbOS send to emit a (size, buffer) pair, rather than Arbitrum value
* Change ArbOS logs to emit a (size, buffer) 2-tuple everywhere there was previously a byte stack
* Change ArbOS input messages to change the type-specific data from a single byte stack field to two fields, size and buffer
* Change how tx id is computed for unsigned EOA transactions, to use keccak256 of data, rather than byte stack hash.
* Conforming changes to Rust front-end, tests, and test logs


hkalodner (approved):
LGTM

PR #317, opened 2021-01-14
Sideload

edfelten:
* Change `sideload` AVM instruction to take a stack argument, which is meant to be a block number.
* Change ArbOS to execute `sideload` with the current block number, when it knows there will be no more naturally occurring incoming requests at that block number.  If `sideload` returns non-empty result, it's treated as an incoming request that is to be processed.
* Add an `adminMode` field to incoming requests. If true, this exempts the request from the gas auction and allows it to run as soon as possible. `adminMode` is false for ordinary requests, and set to true for requests that arrive via `sideload`.

The purpose of this is to make it easy for an aggregator to insert non-mutating calls after executing an L1 block.  In normal, on-chain execution, `sideload` will always return empty. But an aggregator can use the execution as `sideload` as an opportunity to insert (normally non-mutating) calls into the chain for private execution.


PlasmaPower:
This should probably be merged since https://github.com/OffchainLabs/arbitrum/pull/764 is

arb_os/inbox.mini
```diff
@@ -157,24 +157,57 @@ public impure func inbox_get() -> IncomingRequest {
...
+                        aggregator: None<AggregatorInfo>,
+                        adminMode: true,
+                    };
+
+                    if (newMsg.blockNumber == 0) {
+                        newMsg = newMsg with { blockNumber: globalInbox.blockNum };
+                    }
+                    if (newMsg.timestamp == 0) {
+                        newMsg = newMsg with { timestamp: globalInbox.timestamp };
+                    }
```
hkalodner, line 195: Seems like there's a lot of duplication between here and the sideload block below. Would be good to refactor into a common function probably, possible a function to convert `IncomingRequestFromInbox` to `IncomingRequest` could be used in four different places in this file

arb_os/inbox.mini
```diff
@@ -157,24 +157,57 @@ public impure func inbox_get() -> IncomingRequest {
...
                     timestamp: ebMsg.timestamp
                 } with {
                     seenMsgAtThisBlockNum: true
                 };
             }
+        } else {
+            globalInbox = globalInbox with { queue: updatedQ };
         }
 
         globalInbox = globalInbox with { queue: updatedQ };
```
hkalodner, line 213: It seems like the `globalInbox` is being updated twice in all the code paths that reach this point. Maybe it can just be updated here and not in the other places?

hkalodner (changes requested):

arb_os/inbox.mini
edfelten: Factored this out into a common function.

arb_os/inbox.mini
edfelten: Removed the redundant assignment.

hkalodner (approved):
LGTM

PR #319, opened 2021-01-17
Remove old token code

edfelten:
Remove code that implemented the old token model, which we're not using anymore.


hkalodner (approved):
LGTM

PR #320, opened 2021-01-18
Add type byte to beginning of AVM send contents

edfelten:
Add a send-type byte to the beginning of the data emitted on an AVM send.  Type 0 is a standard outbox batch.


hkalodner (approved):
LGTM

PR #323, opened 2021-01-19
If expressions

RGates94:
This PR creates the initial implementation of if expressions. For parsing reasons, these currently use the `xif` keyword to distinguish them from the `if` statement.

This PR moves towards removal of code related to the `IfArm` struct and tracking for early returns.


edfelten (approved):
Looks good.  Eager to try this.

PR #326, opened 2021-01-21
If let expressions

RGates94:
This adds if let expressions to the language, and allows both if let and if expressions to be chained using elseif, including mixtures in the same chain.

This will allow for various code simplifications once all terminal `if`-`if let` statements are replaced by `return` statements using `if`-`if let` expressions

### DIFF
--- a/src/compile/typecheck.rs
+++ b/src/compile/typecheck.rs
@@ -643,7 +643,8 @@ impl AbstractSyntaxTree for TypeCheckedExpr {
                 TypeCheckedNode::Expression(exp2),
                 TypeCheckedNode::Expression(exp3),
             ],
-            TypeCheckedExprKind::If(cond, block, else_block, _) => cond
+            TypeCheckedExprKind::If(cond, block, else_block, _)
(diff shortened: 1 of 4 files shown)


src/compile/typecheck.rs
```diff
@@ -646,6 +654,7 @@ impl AbstractSyntaxTree for TypeCheckedExpr {
                         .flatten(),
                 )
                 .collect(),
+            TypeCheckedExprKind::IfLet(_, _, _, _, _) => unimplemented!(),
```
edfelten, line 657: Need a comment here to explain what's up.  Is this something that needs to be implemented, before if let expressions can be used?  Or is this something that should never be called, for some reason that can be explained?

edfelten (changes requested):
Asking for a clarifying comment in one place.

src/compile/typecheck.rs
RGates94: This logic is currently only used by the inlining code, so I didn't notice it still needed to be implemented when I submitted this PR.

Fixing this will only need a couple lines so expect to see a fix soon.

edfelten (approved):
Looks good.

PR #328, opened 2021-01-22
Remove unused function used to work around old compiler bug

RGates94:
This function is no longer necessary


edfelten (approved):
Good catch.

PR #329, opened 2021-01-22
Simplify type check

RGates94:
Simplifies control flow and removes unneeded code in typecheck.rs.

### DIFF
--- a/src/compile/typecheck.rs
+++ b/src/compile/typecheck.rs
@@ -1006,11 +1006,7 @@ fn typecheck_statement<'a>(
                 type_tree,
                 scopes,
             )?;
-            if return_type.get_representation(type_tree)?.assignable(
-                &tc_expr.get_type().get_representation(type_tree)?,
-                type_tree,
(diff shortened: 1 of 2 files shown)


src/compile/typecheck.rs
```diff
@@ -17,7 +17,7 @@ use crate::stringtable::{StringId, StringTable};
 use crate::uint256::Uint256;
 use std::collections::{HashMap, HashSet};
 
-type TypeTable<'a> = HashMap<usize, &'a Type>;
+type TypeTable<'a> = HashMap<usize, Type>;
```
edfelten, line 20: Do we really need the <`a> on TypeTable?

edfelten (changes requested):
One tiny issue.

src/compile/typecheck.rs
RGates94: Good catch, its no longer needed.

edfelten (approved):
Looks good.

PR #331, opened 2021-01-25
Simple return path

RGates94:
This PR makes the following changes:

- Produces an error if the final statement of a non-void function is not a return statement
- Allows the use of loop and panic as expressions of type `Every`
- Porting of ArbOs and other mini code to use this new syntax
- Changes internal representation of `if` and `if let` statements to use `if` and `if let` expressions.
- Remove old code for generating `if` and `if let` statements
- Fixes issue in `mavm_codegen_code_block` where extra `noop` instructions are generated.


edfelten (approved):
Looks good.

PR #333, opened 2021-01-28
Fix number of locals in codegen of return expression of CodeBlock

RGates94:
Fixes bug in title, makes failing test pass.


edfelten (approved):
Looks good.

PR #334, opened 2021-01-31
Blake2 precompile

edfelten:
* Add support (in Rust AVM) for a new `blake2f` instruction, at opcode 0x26. The instruction takes a single stack argument, which must be a buffer. It reads 213 bytes from the buffer, zeroes the first two bytes, ands the last byte with 0x01, then does the equivalent of the Ethereum 0x09 precompile. This produces a 64-byte output, which is copied into a new buffer, which is left on the stack as the instruction's result.
* Add support for Ethereum's Blake2 precompile, at address 0x09. We limit the number of rounds in this call to 65,535, so the core of the operation can be implemented with the `blake2f` instruction.
* Add tests for this functionality, using the test vectors from EIP-152.


edfelten:
Converted this to draft until we decide our strategy for dealing with L1 gas limits in the one-step proof.

hkalodner (approved):
LGTM though I haven't independently verified the correctness of the hash function implementation

PR #335, opened 2021-02-01
Fix type bug

edfelten:
Fix incorrect type declaration for globalInboxQueue.txQ.  It was declared as an `IncomingRequestQueue` but actually it holds `TxRequestData` values.


michaelsproul:
I think the use here also needs to be updated: https://github.com/OffchainLabs/arb-os/blob/f80d8bd1045f43a1c8f56116354ddbea5610b517/arb_os/inbox.mini#L473

edfelten:
> I think the use here also needs to be updated:
> 
> https://github.com/OffchainLabs/arb-os/blob/f80d8bd1045f43a1c8f56116354ddbea5610b517/arb_os/inbox.mini#L473

Thanks.  Fixed this.

hkalodner (approved):
LGTM

PR #336, opened 2021-02-01
Fix bug in handling of transfers to self

edfelten:
ArbOS previously mishandled transfers of ETH to oneself. 
* Fix this problem by checking for self-transfers in `accountStore_transferEthBalance`
* Add a regression test


hkalodner (approved):
LGTM

PR #337, opened 2021-02-02
Fix bug that caused panic if constructor reverted

edfelten:
Fix a bug that was causing a revert message to be truncated, if a constructor invoked another constructor and the inner constructor reverted.

The fix removes a bunch of fiddly assembly code, so it simplifies the handling of the CREATE and CREATE2 EVM opcodes.


src/evm/mod.rs
```diff
@@ -1865,6 +1865,37 @@ pub fn evm_direct_deploy_and_compressed_call_add(log_to: Option<&Path>, debug: b
...
+fn evm_reverter_factory_test() {
+    _evm_reverter_factory_test_impl();
+}
+
+fn _evm_reverter_factory_test_impl() {
+    let rt_env = RuntimeEnvironment::new(Uint256::from_usize(1111), None);
+    let mut machine = load_from_file(Path::new("arb_os/arbos.mexe"), rt_env);
+    machine.start_at_zero();
+
+    let _contract = match AbiForContract::new_from_file(&test_contract_path("ReverterFactory")) {
```
hkalodner, line 1878: I don't see a `ReverterFactory` contract in the test package. Should that exist?

hkalodner (changes requested):

src/evm/mod.rs
edfelten: Added it.

hkalodner (approved):
LGTM

PR #338, opened 2021-02-04
Simplify contracts build

hkalodner:
- Cut a bunch of unnecessary dependencies from contracts modules
- Update compiled arbos
- Update replay tests


edfelten (approved):
Looks good.

PR #339, opened 2021-02-04
Compatibility fixes

edfelten:
* Swap the order of sub-fields in the `ecpairing` instruction, to match the AVM spec
* Increment the log counter when issuing a log-before-send
* Update test log files due to change in log counter


hkalodner (approved):
LGTM

PR #341, opened 2021-02-08
Mexe reformat

RGates94:
Adds reformat subcommand that changes the format of an existing mexe file without recompiling it.


src/mavm.rs
```diff
@@ -186,14 +186,8 @@ impl Instruction {
...
         match &self.immediate {
-            Some(v) => match self.debug_info.location {
-                Some(loc) => write!(f, "[{}] {}\t\t{}", v, self.opcode, loc),
-                None => write!(f, "[{}] {}\t\t[no location]", v, self.opcode),
-            },
-            None => match self.debug_info.location {
-                Some(loc) => write!(f, "{}\t\t{}", self.opcode, loc),
-                None => write!(f, "{}", self.opcode),
-            },
+            Some(v) => write!(f, "{} [{}]", self.opcode, v),
```
edfelten, line 189: It seems more natural to put the immediate before the opcode, because the immediate will be pushed onto the stack before the opcode executes.

edfelten (changes requested):
Looks good.  One suggestion.

edfelten (approved):
Looks good.

PR #342, opened 2021-02-11
Mexe remove virtual

RGates94:
Removes the ability for a `LinkedProgram` to store virtual opcodes, thus preventing them from being written to a mexe, and reducing their size.  Also limits emulator to running on `AVMOpcode`s, in order to reduce the number of error cases and reduce boilerplate.  As a side effect, `Instruction` is now generic over the `opcode` field, with `Opcode` as a default.


edfelten (approved):
Looks good.

PR #347, opened 2021-02-12
Remove add_auto_link_progs

RGates94:
Removes add_auto_link_progs, which adds to the size of the mexe unnecessarily.


edfelten (approved):
Looks good.  Always happy to remove unneeded code.

PR #349, opened 2021-02-13
Don't advance the inbox queue unless we're consuming a message

### LINKED ISSUE #343: Incoming messages sometimes dropped in sequencer mode
In inbox.mini, the queue is advanced (but shouldn't be) when this if statement is true: `if ((globalInbox.sequencer != None<SequencerState>) && globalInbox.seenMsgAtThisBlockNum)`

The result is that a message can be dropped.

edfelten:
The previous code sometimes advanced the inbox queue (discarding the message at its head), in a case where that message wouldn't be consumed.  This PR moves the code that advances the queue to a place where we know the head message will definitely be consumed.

Fixes #343.


RGates94 (approved):
Looks good.

PR #350, opened 2021-02-13
Restrict non eip 155 signatures

edfelten:
Restrict the types of transactions that can be accepted if signed by a non-EIP-155 signature. In particular, we reject the signature on a transaction unless it is a deploy transaction, and has sequence number (nonce) of zero, and has zero callvalue.

The purpose of the restriction is to avoid attacks that replay Ethereum transactions from the pre-EIP-155 era. In particular, we don't want to allow old payment transactions to be replayed.

We do allow such signatures in a restricted case, to ensure that keyless deployments will still work.

This PR also refactors the TxRequestData struct.

### DIFF
--- a/arb_os/decompression.mini
+++ b/arb_os/decompression.mini
@@ -424,8 +424,7 @@ impure func decompressTx(
             calleeAddr: toAddr,
             value: value,
             calldata: calldata,
-            nonMutating: false,
-            isConstructor: (toAddr == address(0)),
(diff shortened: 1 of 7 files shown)


hkalodner:
Personally I'm against this change since it seems to be compatibility breaking to me. Assuming the matic and xdai allow pre eip155 sigs, I think we should too. If I'm wrong and they don't, I would support this change

edfelten:
Closing this because the underlying issue is WONTFIX.

PR #352, opened 2021-02-19
Fix typecheck bug with casts

### LINKED ISSUE #116: Constant folding gives wrong answer for ToAddress
In the compiler, constant folding of a expression like `address(someUint)` should truncate the value of `someUint` to 160 bits.  Currently it doesn't do any masking.

An example expression that yields the wrong result is `address(hash(0))`

RGates94:
Fixes a bug where casts like `uint`, `int`, `bytes32`, and `address` casts did not check if the subexpression is of the correct type if it was a constant. For example, previously, the statement `let x = address(None<uint>);` would compile.

This also fixes #116.

### DIFF
--- a/src/compile/typecheck.rs
+++ b/src/compile/typecheck.rs
@@ -2134,8 +2134,8 @@ fn typecheck_unary_op(
             )),
         },
         UnaryOp::ToUint => {
-            if let TypeCheckedExprKind::Const(val, _) = sub_expr.kind {
-                Ok(TypeCheckedExprKind::Const(val, Type::Uint))
+            if let TypeCheckedExprKind::Const(Value::Int(val), _) = sub_expr.kind {
+                Ok(TypeCheckedExprKind::Const(Value::Int(val), Type::Uint))
             } else {
                 match tc_type {
                     Type::Uint | Type::Int | Type::Bytes32 | Type::EthAddress | Type::Bool => {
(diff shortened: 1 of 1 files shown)


src/compile/typecheck.rs
```diff
@@ -2187,8 +2187,19 @@ fn typecheck_unary_op(
...
         UnaryOp::ToAddress => {
-            if let TypeCheckedExprKind::Const(val, _) = sub_expr.kind {
-                Ok(TypeCheckedExprKind::Const(val, Type::EthAddress))
+            if let TypeCheckedExprKind::Const(Value::Int(val), _) = sub_expr.kind {
+                Ok(TypeCheckedExprKind::Const(
+                    Value::Int(
+                        val.modulo(
+                            &Uint256::from_string_hex(
+                                "1__0000_0000__0000_0000__0000_0000__0000_0000__0000_0000",
+                            ) //2^40, 1+max address
```
edfelten, line 2196: Isn't this really 2^160?

edfelten (changes requested):
Just one apparent error in a comment.  Otherwise good to go.

src/compile/typecheck.rs
RGates94: Yes, I'll fix that.

edfelten (approved):
Looks good.

PR #354, opened 2021-02-24
Hk fixes

hkalodner:
- Change `gracePeriodBlocks` to expect config parameter as blocks instead of ticks [note: This might require changes to other tests, but nothing that I saw]
- Output send merkle root log before block log
- Fix merklization of sends
- Fix send block statistics


edfelten (approved):
Looks good, after a couple of fixes to the test harness.

PR #356, opened 2021-02-25
Merge match patterns

RGates94:
Merges the definitions of MatchPattern and TypeCheckedMatchPattern via use of generics.  This is necessary to effectively merge type checked versions of more complex ast nodes, and reduces the number of locations one needs to change to edit the match pattern definition.


edfelten (approved):
Looks good.

PR #357, opened 2021-02-25
Merge struct initializers and funcs

RGates94:
Merges `FieldInitializer` with `TypeCheckedStructField`, and `FuncDecl` with `TypeCheckedFunc`, reducing complexity and duplication of fields.


edfelten (approved):
Looks good.

PR #359, opened 2021-02-27
Delay non-tx-receipt logs until end of block

edfelten:
This changes how AVM logs are emitted. Any AVM logs, other than tx receipts, will be delayed until the end of the Arbitrum block, and released at the end of the block, just before the block summary log.

This ensures that tx receipts are always contiguous at the beginning of a block's logs.


hkalodner (approved):
LGTM, tested against monorepo and updated tests in https://github.com/OffchainLabs/arbitrum/pull/783

PR #360, opened 2021-02-27
Use Arbitrum block numbers in addition to Ethereum block numbers

edfelten:
Add a notion of Arbitrum block number.  An Arbitrum block is a subset of an Ethereum block. The two are numbered separately.  For more information on Arbitrum blocks see the separate document explaining how block numbering works.


arb_os/output.mini
```diff
@@ -116,17 +113,10 @@ public impure func outputStats_endOfBlock(blockNum: uint, timestamp: uint) {
...
-        let gasLimit = chainParams_speedLimitPerBlock();
-        if (gasLimit < gasUsedThisBlock) {
-            gasLimit = gasUsedThisBlock;
-        }
-
         asm((
             const::LogType_blockSummary,
             blockNum,
             timestamp,
-            gasLimit,
```
hkalodner, line 129: The RPC interface is gonna need to use something as the block gas limit. Not sure what makes sense for that

arb_os/arbsys.mini
```diff
@@ -276,7 +290,7 @@ impure func arbsys_sendTxToL1(topFrame: EvmCallFrame, calldata: ByteArray) { //
                                 uint(destAddr),
                             ),
                             96,
-                            inbox_currentBlockNumber(),
+                            inbox_currentArbBlockNumber(),
```
hkalodner, line 293: Might want to also output the eth block number for this function

arb_os/inbox.mini
```diff
@@ -228,174 +150,72 @@ impure func incomingRequestFromRaw(newMsgRaw: IncomingRequestFromInbox, adminMod
...
-    } else {
-        seq = seq with { lastBlockNumSeen: blockNum };
-    }
-    let timestamp = newMsg.timestamp;
-    if (timestamp == 0) {
-        timestamp = seq.lastBlockNumSeen;
-    } else {
-        seq = seq with { lastTimestampSeen: timestamp };
+impure func endOfArbitrumBlock(inbox: Inbox) -> Inbox {
+    let rawSideloadMsg = asm(inbox.arbBlockNum,) any { sideload };
```
hkalodner, line 190: It seems like this will make the side loaded message run in the next arbitrum block rather than the current one since it increments `arbBlockNum`

arb_os/gasAccounting.mini
```diff
@@ -785,28 +787,31 @@ impure func payForTxAndL1Gas(
...
     let congestionPrice = state.congestionPriceWei;
     let basePrice = gasAccountingInfo.currentPrices.perArbGasBase;
     let gasPool = state.gasPool;
-    let speedLimit = gasAccountingParams.SpeedLimitPerBlock;
-    while((blocksToAdvance > 0) && ( (state.gasPool < gasAccountingParams.GasPoolMax) || (congestionPrice > 0) )) {
-        blocksToAdvance = blocksToAdvance-1;
+    let gasPoolMax = gasAccountingParams.GasPoolMax;
+    let speedLimit = gasAccountingParams.SpeedLimitPerSecond;
+
+    //TODO: solve this analytically, so we don't have to do a per-second loop
```
hkalodner, line 800: Noting the TODO here

hkalodner (changes requested):

arb_os/output.mini
edfelten: A good option is 13.5 times the per-second speed limit.  That will be correct on average, on a busy non-sequencer chain.

doc/DataFormats.md
```diff
@@ -54,8 +54,8 @@ This message type is initiated by the EthBridge, as part of the creation of a ne
 
 Type-specific data:
 
-* challenge period, in L1 blocks (uint)
-* ArbGas speed limit, in ArbGas per L1 block (uint)
+* challenge period, in seconds (uint)
+* ArbGas speed limit, in ArbGas per second (uint)
```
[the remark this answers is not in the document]
edfelten: At the moment the "challenge period, in second" parameter isn't used for anything in ArbOS.  We could make it be measured in blocks, but my guess was that seconds would more likely be the unit at whatever future point this might start getting used.

arb_os/arbsys.mini
edfelten: Added the Eth block number, immediately after the Arbitrum block number.

arb_os/inbox.mini
edfelten: I don't think so.  The incoming message is assigned its arbBlockNum by `incomingRequestFromRaw` which runs before `inbox.arbBlockNum` gets incremented.

arb_os/gasAccounting.mini
edfelten: I updated the comment, and tightened up the code a bit.  I think the cost of this is minimal, maybe a few hundred ArbGas per second at most.  100 ArbGas is 0.0001% of the chain's capacity.

arb_os/inbox.mini
hkalodner: Makes sense

arb_os/gasAccounting.mini
hkalodner: Haha sounds good

hkalodner (changes requested):
LGTM other than needing an updated arbos.mexe

hkalodner (approved):
LGTM

PR #361, opened 2021-03-01
Make BLOCKHASH depend on chain history

edfelten:
This changes the EVM BLOCKHASH instruction so that it returns a value that depends on the history of the chain.  Specifically, the blockhash for an Ethereum block is hash(ethBlockNum, inboxAccum) where inboxAccum is a hash accumulator of all inbox messages up to and including block number ethBlockNum.


hkalodner (approved):
LGTM

PR #363, opened 2021-03-01
Fix loop termination bug

edfelten:
Fix bug in loop termination condition. This was causing the loop to run for too long, which didn't affect results but could waste considerable resources.


hkalodner (approved):
LGTM

PR #365, opened 2021-03-02
Improve fee stats in tx receipt

edfelten:
Change the fee statistics field in tx receipts to provide more information.  New format is documented in the data formats document.


hkalodner (approved):
LGTM

PR #367, opened 2021-03-03
Fix typo and expect speed limit param to be in blocks

edfelten:
* Fix typo in chainParameters
* Expect the chain init message to give ArbGas speed limit in blocks rather than seconds


hkalodner (approved):
LGTM

PR #370, opened 2021-03-03
Fix struct typecheck

RGates94:
Fixes typechecking bug where structs with differently named but identically typed fields could be assigned to each other.  Fixes occurrences of this behavior in the mini code.


arb_os/gasAccounting.mini
```diff
@@ -188,7 +188,7 @@ public impure func setGasAccountingParams(speedLimitPerBlock: uint, gasPoolMax:
         false
     } else {
         gasAccountingParams = struct {
-            SpeedLimitPerBlock: speedLimitPerBlock,
+            SpeedLimitPerSecond: speedLimitPerBlock,
```
edfelten, line 191: Units don't match here.  We can't set a perSecond variable to a perBlock value.  I'll push a commit to fix this.

edfelten (changes requested):
One problem, which I'll fix right away.

arb_os/gasAccounting.mini
edfelten: Fixed this.

edfelten (approved):
Looks good, as modified.

PR #371, opened 2021-03-03
Update eth tooling

fredlacs:
This PR changes eth tooling to use hardhat instead of truffle to compile contracts.
To do so the folder structure was refactored and the corresponding file paths in the Rust code updated.


edfelten (approved):
Looks good.

PR #375, opened 2021-03-04
Preferred and default aggregator support

edfelten:
* Add a chain parameter that can specify a default aggregator, settable as a chain init option or via a builtin contract call
* Add a preferred aggregator attribute to L2 accounts. This can be an aggregator address, or None to specify using the default aggregator. The attribute is initially None for all accounts, and can be modified via a builtin contract call.
* When charging fees to a tx, only reimburse the validator if the tx was submitted by the sender's preferred validator (or the default validator if their preferred validator is None.) Otherwise, don't charge the sender for the reimbursable portion of fees.
* Add an ArbAggregator builtin contract to allow getting and setting of preferred aggregators (settable only by the account itself) and the default aggregator (settable only by the chain owner or the current default aggregator).
* Add tests

This functionality prevents an attack where a malicious aggregator front-runs an honest aggregator's transaction submissions in order to capture the honest aggregator's reimbursements.

If a chain has a single aggregator that is used by most users by default, that should be set as the default aggregator, so that users don't have to configure anything before using that aggregator.


node_modules/.yarn-integrity
```diff
@@ -0,0 +1,12 @@
+{
```
hkalodner, line 1: This file should be removed

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
hkalodner, line 1: This file should be removed

src/compile/miniconstants.rs
```diff
@@ -19,6 +19,8 @@ pub fn init_constant_table() -> HashMap<String, Uint256> {
         ("Address_ArbosTest", 105),
         ("Address_ArbOwner", 107),
         ("Address_ArbGasInfo", 108),
+        ("Address_ArbAggregator", 109),
+        // address 110 reserved for EthBridge functionality
```
hkalodner, line 23: If we're reserving this, I think the idea was to make any real call to the L2 contract at that address revert

arb_os/gasAccounting.mini
```diff
@@ -798,7 +808,12 @@ impure func payForTxAndL1Gas(
         fullAmount
     };
 
-    let payableToFee = fixedPoint_mulByUintTrunc(feeConfig.shareForNetFee, amount);
+    let payableToFee = xif(preferredAggregator == recipient) {
+        fixedPoint_mulByUintTrunc(feeConfig.shareForNetFee, amount)
+    } else {
+        amount
+    };
```
hkalodner, line 815: I'm not sure I understand why we're only incorporating the network fee here if using the preferred aggregator. I assume is has to do with the adjustment we made if you're not using the preferred aggregator above, but this is confusing and needs some refactoring or documentation

hkalodner (changes requested):

arb_os/chainParameters.mini
```diff
@@ -29,6 +29,7 @@ type ChainParams = struct {
     stakingToken: address,           // ERC-20 token used for staking; or zero if using ETH
     owner: address,                  // owner of the chain, who has admin privileges (or zero if there is no owner)
     secondsPerSend: option<uint>,    // issue at most one send per this many seconds
+    defaultAggregator: address,      // default aggregator, for users who haven't specified one
```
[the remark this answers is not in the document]
edfelten: This uses zero as the default aggregator address.  (`option` doesn't work well with Solidity APIs.)

node_modules/.yarn-integrity
edfelten: Removed

src/compile/miniconstants.rs
edfelten: Added code to do this.

yarn.lock
edfelten: Done.

arb_os/gasAccounting.mini
edfelten: Added comments to explain this.

hkalodner (approved):
LGTM

PR #376, opened 2021-03-04
Add ArbSys call to check whether you were directly called from L1

edfelten:
Add an ArbSys method, `calledFromL1` which returns true iff its caller was called directly from L1.

It is tempting but incorrect to detect this condition using `msg.sender == tx.origin` but that could be incorrect if `tx.origin` is an L1 contract and `msg.sender` is an L2 contract at the same address.


contracts/arbos/builtin/ArbSys.sol
```diff
@@ -22,6 +22,9 @@ interface ArbSys {
     // able to call it
     function getStorageAt(address account, uint256 index) external view returns (uint256);
 
+    // Return true if the caller of this was called directly from L1
+    function calledFromL1() external view returns (bool);
```
hkalodner, line 26: I think this would be better framed as `isTopLevel` or `isTopLevelCall`. Other than this this looks good to me

hkalodner (changes requested):

contracts/arbos/builtin/ArbSys.sol
edfelten: Renamed to `isTopLevelCall`.

hkalodner (approved):
LGTM

PR #379, opened 2021-03-04
Fix linking bug

RGates94:
Fixes bug that was causing the structure of the register to be different than expected.


edfelten (approved):
Good catch.

PR #380, opened 2021-03-05
EVM log on l2 to l1 call

edfelten:
This makes ArbSys emit an event (EVM log) on L2 to L1 calls.


edfelten:
> Should we maybe use a more standard format to set this topic?
> I'd suggest something like `Keccak256(L2ToL1Tx(address,bytes32))` which would result in `3ea427d374f4e940e85a46ed68102f8b1064a73ad8472a9ae151c1fd3ddb0c92`
> Address being the destAddress and bytes32 a hash of the calldataForL1

We do use the standard topic schema for the emitted logs.  (The nonstandard placeholder is only used internally, to mark the item as needing some post-processing.  When the post-processing occurs, the standard-schema topic is substituted in.)

arb_os/output.mini
```diff
@@ -343,14 +344,61 @@ public impure func queueMessageForSend(msg: ByteArray) {
...
+impure func updateEvmLogForSendToL1(rawLog: any, state: any) -> (any, any) {
+    // This is designed to be used by evmlogs_forall. When used by evmlogs_forall in its pass through
+    //      the current set of logs, this morphs the first EVM log with topic[0] == const::EVMLogTopicPlaceHolderForL2ToL1Send
+    //      into a proper EVM log item with batch-related fields filled in
+
+    return xif(unsafecast<bool>(state)) {
+        // Cast the log item to a struct that assumes 3 topics. topic1 and topic2 might not exist,
+        //      but that's OK because we won't try to access them until after we have verified that they exist.
+        let log = unsafecast<LogWith3Topics>(rawLog);
+        xif (log.topic0 == const::EVMLogTopicPlaceHolderForL2ToL1Send) {
```
hkalodner, line 388: I think 0 topic logs will break this. Can we check the tuple length first to make sure it's big enough?

arb_os/output.mini
```diff
@@ -343,14 +344,61 @@ public impure func queueMessageForSend(msg: ByteArray) {
     evmCallStack_queueMessage(msg);
 }
 
-public impure func sendQueuedMessages(queue: Queue) {
-    loop {
+public impure func sendQueuedMessages(queue: Queue, evmLogs: EvmLogs) -> EvmLogs {
+    return xloop {
```
hkalodner, line 348: I don't love the fact that we're using an O(n^2) algorithm loop for this. If this get's charged to overhead, that could potentially be an issue

hkalodner (changes requested):

hkalodner (changes requested):
One other issue. This PR makes a breaking change to reduce the typecode length in the sendTxToL1 virtual send to one byte which I support. However it doesn't adjust the typecode length in the other two virtual send types. I think both should be updated similarly to maintain compatibility. The other option is of course removal which is planned for buddy contracts, and probably should be done for pure eth withdrawals as well

arb_os/output.mini
edfelten: Added a length check.

edfelten:
> One other issue. This PR makes a breaking change to reduce the typecode length in the sendTxToL1 virtual send to one byte which I support. However it doesn't adjust the typecode length in the other two virtual send types. I think both should be updated similarly to maintain compatibility. The other option is of course removal which is planned for buddy contracts, and probably should be done for pure eth withdrawals as well

I made this change, to make the type field in sends 1 byte rather than 32.  (The documentation already said it is 1 byte.)

arb_os/output.mini
edfelten: Good catch. Switched to an O(n) algorithm.

hkalodner (changes requested):

arb_os/arbsys.mini
```diff
@@ -284,40 +287,94 @@ impure func arbsys_sendTxToL1(topFrame: EvmCallFrame, calldata: ByteArray) { //
...
+                        4*32,
+                        uint(destAddr),
+                    ),
+                    5*32,
+                    callvalue,
+                ),
+                6*32,
+                0xe,
+            ),
+            7*32,
```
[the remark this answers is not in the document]
edfelten: This is what the usual online ABI formatter did.  It did the bytes as: 0xe (32 bytes); size (32 bytes); data.  Do you know what is correct?

src/compile/miniconstants.rs
```diff
@@ -239,6 +239,11 @@ pub fn init_constant_table() -> HashMap<String, Uint256> {
             // Keccak256 of "Arbitrum BLS signature domain"
             "73a92f91d473214defd5ffa91d036007eb2e6487fffaa551835e988fb24aaa2b",
         ),
+        (
+            "EVMLogTopicForL2ToL1Send",
+            // Keccak256 of "Arbitrum EVM Log Placeholder for L2 to L1 Send"
+            "1389380ee1d43ae62f988fd47b8d5e9fc263444ea53d0cc330fc4869d7a8689a"
```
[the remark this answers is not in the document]
edfelten: Hmm, not sure why I had the wrong hash here.  Fixed now.

arb_os/arbsys.mini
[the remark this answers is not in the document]
edfelten: I think this is fixed now.

hkalodner (approved):
LGTM

PR #384, opened 2021-03-08
Adding a field to track maximal modified index for buffers

mrsmkl:


src/run/emulator.rs
```diff
@@ -1324,6 +1357,9 @@ impl Machine {
                 } else {
                     self.arb_gas_remaining.clone()
                 };
+                if let Some(val) = &insn.immediate {
```
edfelten, line 1360: Why was this code moved?  It seems like it is simpler to push the immediate onto the stack before calculating gas cost, because this saves the gas cost computation from needing to worry about whether the instruction's args come from the immediate or the stack.

Unless there is a reason to move this, I would prefer to keep it in the original place, and adjust the setbuffer gas cost calculation accordingly.

If this stays where it is, the pairing gas cost calculation will need to be fixed.

edfelten (changes requested):
Looks good, except for one issue raised in a comment.

src/run/emulator.rs
mrsmkl: I moved it because there are some places where `next_op_gas` is called and it doesn't look like the immediate is pushed into stack before those ones.

In `gas_for_pairing`, it gets the stack value using `self.stack.contents.get(0)`, but looks like this is the bottom of the stack, not the top?

Perhaps it would be best to pass the top elements of the stack to the functions for variable gas.

edfelten (approved):

edfelten:
Approved this. Holding off on merging until the matching PR in the Arbitrum repo is ready.

PR #386, opened 2021-03-09
Eliminate withdrawETH L2 to L1 message

edfelten:
Eliminate this message, which is no longer needed. Use L2-to-L1 transaction send instead.


hkalodner (dismissed):
LGTM after rebase

PR #387, opened 2021-03-09
Expand unsigned tx messages

edfelten:
Add fields to unsigned tx messages from contracts, to allow the sender to make gas money refundable to a 3rd party if not used, and to allow sender to make callvalue refundable to a 3rd party if the tx reverts. This provides flexibility useful in some L1 to L2 bridging use cases.

(Actually, instead of handling this as a refund, this code transfers the funds from the sender to the "refund recipient",  then makes the "refund recipient" pay for the gas or callvalue. This has the same effect but is more natural given how ArbOS handles gas and callvalue.)

This PR includes all of the changes in #385


edfelten:
Closing this because similar functionality was included in a different PR.

PR #389, opened 2021-03-09
Add builtin ArbStatistics contract making available chain statistics

edfelten:


arb_os/evmCallStack.mini
```diff
@@ -783,6 +785,8 @@ public impure func evmCallStack_returnFromCall(
                             Some(gasUsage),
                             feeStats,
                         );
+
+                        arbStatistics_addContract();
```
hkalodner, line 789: This covers externally created contracts, but does it cover `CREATE` and `CREATE2`?

arb_os/arbstatistics.mini
```diff
@@ -0,0 +1,103 @@
...
+    } else {
+        evmOp_revert_knownPc(12, 0, 0);
+    }
+}
+
+var arbStats: struct {
+    storage: uint,
+    arbGasUsed: uint,
+    numTxReceipts: uint,
+    numContracts: uint,
```
hkalodner, line 69: It looks like this doesn't take get reduced by `SELFDESTRUCT`. Is that intentional? If so maybe `numContractsCreated` would be better

hkalodner (changes requested):

arb_os/evmCallStack.mini
edfelten: Added coverage for CREATE and CREATE2

arb_os/arbstatistics.mini
edfelten: Added coverage for SELFDESTRUCT

hkalodner (approved):
LGTM

PR #390, opened 2021-03-10
Runtime upgrades for ArbOS

edfelten:
This PR includes an apparently working version of runtime upgrades for ArbOS.  It includes an automatic generator that reads the pre-upgrade and post-upgrade .mexe files and automatically generates code to rearrange the old program's globals and install the new program's jump table.  It also makes minor compiler changes, to provide the affordances ArbOS needs to tell whether it's running as an upgrade and to allow remapping of globals. Finally, it adds an alternate startup path that re-initializes just enough to switch over to the new code without destroying any of the old state.


arb_os/accounts.mini
```diff
@@ -267,23 +267,47 @@ public impure func accountStore_createAccountFromEvmCode(
...
+    let oldAcct = accountStore_get(store, addr);
+    return Some(accountStore_set(
+        store,
+        addr,
+        oldAcct with {
+            contractInfo: Some(struct {
+                code: code,
+                evmJumpTable: evmJumpTable,
+                startCodePoint: initCodePt,
+                storage: (oldAcct.contractInfo?).storage,
```
edfelten, line 286: This should deal with the case where `oldAcct.contractInfo` is None

arb_os/main.mini
```diff
@@ -39,35 +41,52 @@ use std::bytestream::ByteStream;
...
-    txReceipts_init();
-    rollupTracker_init();
-    if (initializePreinstalls() == None<()>) {
-        panic;   // don't try to run without preinstalls and precompiles
+        // now re-do the initializations that set up pointers into the ArbOS code
+        errorHandler_init();
+        evmOps_init();
+        evmCallStack_makeEmpty();   // clear out the old EVM call stack from before the upgrade
+        if (initializePreinstalls(true) == None<()>) {
+            panic;   // don't try to run without preinstalls and precompiles
```
edfelten, line 69: Add a comment saying that this will abort the upgrade.

src/compile/ast.rs
```diff
@@ -293,18 +330,24 @@ impl Type {
...
-            Type::FixedArray(t, size) => format!("[{}]{}", size, t.display()),
+            Type::Array(t) => format!("[]{}", t.display_indented(indent_level)),
+            Type::FixedArray(t, size) => format!("[{}]{}", size, t.display_indented(indent_level)),
             Type::Struct(fields) => {
                 let mut out = "struct {\n".to_string();
+                for _ in 0..indent_level {
+                    out.push_str("    ");
+                }
                 for field in fields {
                     //This should indent further when dealing with sub-structs
```
edfelten, line 346: Remove this comment, because we already do what it is asking us to do.

src/compile/mod.rs
```diff
@@ -174,15 +175,15 @@ impl CompiledProgram {
         int_offset: usize,
         ext_offset: usize,
         func_offset: usize,
-        globals_offset: usize,
+        globals_offset: Vec<GlobalVarDecl>,
```
edfelten, line 178: The variable name `globals_offset` is confusing here. Maybe change it to `globals`?

edfelten (commented):
One substantive thing to change.

RGates94 (commented):
Looks good overall, a few small suggestions.

arb_os/accounts.mini
edfelten: Fixed.

hkalodner (approved):
LGTM

PR #391, opened 2021-03-10
Fix inline

RGates94:
Fixes bugs in the inlining code, currently disables inlining, as applying inlining universally incurs an overall gas cost.  Controls to control when inlining is done will be included in a follow up PR.


edfelten (approved):
Looks good.

PR #394, opened 2021-03-11
Parse error reporting

RGates94:
This makes error messages that occur during parsing more readable.


edfelten (changes requested):
Looks good.  Need rebasing.

edfelten (approved):
Looks good.

PR #396, opened 2021-03-11
Retryable transactions

edfelten:
* Allow L1 contracts to upload transactions to L2 without executing them, leaving them in a Retryable Transaction (RTx) buffer.  This mechanism is designed so the L1 program doesn't have to worry about L2 gas considerations.
* Add a precompiled contract that allows RTxs to be executed. If the tx succeeds, it is removed from the RTx buffer.
* RTxs in the buffer have a timeout. Anyone can pay rent to keep an RTx in the buffer longer. An RTx is removed from the buffer if it ages out, or if it is successfully executed.
* RTxs execute as if called by the original submitter.


hkalodner (approved):
LGTM. Other version of PR was thoroughly reviewed and this PR has been tested for monorepo compatibility. Note that I'm only tested the PR without fee's activated which will execute some currently unused codepaths

edfelten:
Approved.  Only change is update to arbos.mexe

PR #400, opened 2021-03-12
Fix inconsistent capitalization that causes error on Linux

edfelten:


RGates94 (commented):
When running make, the test `logfile_replay_tests` fails unless you first run `make testlogs; cp testlogs/* replayTests/`, otherwise is working on both OSX and linux.

RGates94 (approved):
Looks good, working on both osx and linux

PR #402, opened 2021-03-14
Fix selfdestruct bug

edfelten:
Fixes an error that was causing some instances of EVM SELFDESTRUCT to incorrectly abort the transaction.


hkalodner (approved):
LGTM

PR #404, opened 2021-03-15
Fix gas stipend

edfelten:
Fix the automatic gas stipend logic in ArbOS.  Ethereum txs that transfer balance get an automatic gas stipend of 2300 gas. We offer a larger stipend of 20000 ArbGas in Arbitrum.  But we want to arrange things so that if a caller gives a gas amount that would not allow storage writes in Ethereum, we don't allow storage writes in Arbitrum. The logic for ensuring this was inaccurate. This PR fixes that.


hkalodner (approved):
LGTM

PR #405, opened 2021-03-16
Do not allow assignment of void values in let statement

RGates94:
This prevents a compiler crash caused by assignment of type void in a let statement.


edfelten (approved):
Looks good.

PR #406, opened 2021-03-16
Make gen code order alphabetically

RGates94:
This makes the imports and the assignments of globals ordered alphabetically, previously there was no defined order causing some minor nondeterminism.


edfelten (approved):
Looks good.

PR #408, opened 2021-03-17
Flatten buffers in mexe

edfelten:
Change format for serializing buffers to json. This affects the .mexe format as well as testlog formats. This maintains compatibility with the updated buffer format in the Arbitrum repo, and allows easier cross-repo exchange of tests and diagnosis files.


hkalodner (approved):
LGTM

PR #410, opened 2021-03-17
Allow aggregator to direct its fee revenue to another address

edfelten:
This PR allows an aggregator to control where its fee revenues are paid.  Aggregators might want to direct these fees to cold wallet addresses, which will differ from the hot wallets they're identified by on L1 (and to ArbOS).  
* Add a notion of "fee collection address" that can aggregator can change. ArbOS gives an aggregator's fee revenue to its fee collection address.
* An aggregator's fee collection address is initially set to the aggregator's own address.
* An aggregator's fee collection address can be changed only by the chain owner or by that aggregator's current fee collection address.  (The aggregator itself cannot change this, because we want an aggregator to be able to put its fee revenue outside the reach of its hot wallet address.)


hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #411, opened 2021-03-18
Let retryable submitter try to redeem immediately

edfelten:
This gives the submitter of a retryable tx the option of trying to redeem the tx immediately.


hkalodner (approved):
LGTM

PR #413, opened 2021-03-19
Refactor compile from folder

RGates94:
This separates out each compilation stage into its own function rather than keeping all of the logic in the `compile_from_folder` function.  This will help with testing of each stage, and will make configuration of alternate compilation pipelines easier.

### DIFF
--- a/src/compile/mod.rs
+++ b/src/compile/mod.rs
@@ -332,11 +332,9 @@ pub fn compile_from_folder(
     file_name_chart: &mut BTreeMap<u64, String>,
     inline: bool,
 ) -> Result<Vec<CompiledProgram>, CompileError> {
-    //Parsing step
     let (mut programs, import_map) = create_program_tree(folder, library, main, file_name_chart)?;
-    //Resolution of imports (use statements)
     resolve_imports(&mut programs, &import_map)?;
-    //Conversion of programs `HashMap` to `Vec` for typechecking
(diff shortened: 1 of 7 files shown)


src/compile/mod.rs
```diff
@@ -594,6 +519,97 @@ fn create_type_tree(program_tree: &HashMap<Vec<String>, Module>) -> TypeTree {
         .collect()
 }
 
+fn typecheck_programs(
+    type_tree: &TypeTree,
+    output: Vec<Module>,
```
edfelten, line 524: The name `output` is non-intuitive here. Is there a more descriptive name we could use?

src/compile/mod.rs
```diff
@@ -594,6 +519,97 @@ fn create_type_tree(program_tree: &HashMap<Vec<String>, Module>) -> TypeTree {
...
+            exported_funcs,
+            global_vars,
+            name,
+        ));
+    }
+    Ok(typechecked)
+}
+
+fn codegen_programs(
+    typechecked: Vec<TypeCheckedModule>,
```
edfelten, line 576: Can we use a more descriptive name, like `modules`?

edfelten (changes requested):
Looks good overall.  Just two variable names to possibly change.

edfelten (approved):

PR #414, opened 2021-03-20
Support generating upgrade files

edfelten:
This adds a subcommand to the mini tool, to create a json file that packages up the bytes needed to perform an ArbOS upgrade, in a format convenient for other tools to use.


edfelten:
Leaving this as a draft pending testing on the devote.

hkalodner (approved):
LGTM

Noting here that this PR means that the owner should be an EOA from the perspective of the L2 (could be an L1 smart contract)

PR #415, opened 2021-03-22
Attach logs in retryable txs to the retryable

edfelten:
Previously, when a retryable transaction was redeemed, any EVM logs produced by the retryable would be attached to the receipt of the outer transaction that redeemed the retryable. This changes things so that logs produced by the retryable are attached to the retryable own receipt.

This also enhances testing of returnables.


hkalodner (approved):
LGTM

PR #416, opened 2021-03-23
Add configuration for the cargo deny tool

edfelten:
cargo deny is a tool to support vetting of Rust dependences for things like licensing. This PR adds a configuration for cargo deny.


RGates94 (approved):
Looks good

PR #417, opened 2021-03-24
Support gas estimation

edfelten:
Ethereum charges different prices for calldata bytes depending on whether a byte is zero (4 gas) or non-zero (16 gas). This PR changes Arbitrum to use the same policy.  Calldata usage is now accounted for in units, where a zero byte is 4 units and a non-zero byte is 16 units.  This is done consistently through the accounting and reporting in transaction receipts.


hkalodner (approved):
LGTM. Incorporated into arbitrum repo via https://github.com/OffchainLabs/arbitrum/pull/1041

PR #418, opened 2021-03-24
Start new segment with zero opcode

edfelten:
Start each new code segment with a zero opcode, rather than a Panic opcode. This affects the hash of CodePoints.


RGates94 (approved):
Looks good.

PR #420, opened 2021-03-25
Fix handling of zero-gas transactions

edfelten:
* For all txs to contract addresses, require a minimum of 5000 ArbGas to be supplied, otherwise fail the tx with a new error code.
* Clean up evmCallStack when a call completes, so we don't issue a bogus tx receipt if execution errors between txs.
* Add test for low-gas contract tx


hkalodner (approved):
LGTM

hkalodner (dismissed):
LGTM

hkalodner (approved):
LGTM

PR #421, opened 2021-03-26
Runtime environment simplifications

RGates94:
This PR simplifies usage of the `RuntimeEnvironment` struct


edfelten (approved):
Looks good.

PR #422, opened 2021-03-30
Evm module refactor

RGates94:
This PR separates out the tests in the evm module in the mini compiler into their own files, and makes a few simplifications to code in that module.


edfelten (approved):
Looks good.

PR #423, opened 2021-03-30
Sequencer

edfelten:
* Change ArbOS to handle the block number / timestamp behavior of sequencer chains.  ArbOS adjusts the block number and timestamp of each arriving message to ensure that (post-adjustment) the block numbers and timestamps of messages are non-decreasing.
* Refactor the Rust runtime and test harness to allow testing of sequencer chain behavior.


hkalodner (approved):
LGTM

PR #425, opened 2021-04-01
De-duplicate contract code storage

edfelten:
This refactors how the EVM and AVM code for contracts is stored, to save memory by de-duplicating.  The idea is that when two contracts have the same deployed EVM code, they can share a single copy of that EVM code and of the AVM code translated from it. (Studies of Ethereum mainnet suggest that this kind of duplication is very common.)

Account stores now include a map from EVM code hash to relevant values. Items in the table are reference-counted, and dropped when their reference count reaches zero.  The reference count is incremented for each new contract added to the account store that uses the code, and decremented if one of the contract self-destructs.

The reference count and table management are safe, in the sense that an inaccurate reference count doesn't affect correctness of execution, but it could lead to less efficient use of memory because of avoidable duplication of code.

Reference counting and deduplication are done only for deployed code, and not for constructor code, because ArbCore already deals with constructor code fairly efficiently.  If we decide later that we need to de-duplicate constructor code, we could add that capability, but this PR doesn't do that, for simplicity.


hkalodner (approved):
LGTM

PR #426, opened 2021-04-02
Emit info for call tracing

edfelten:
* Add basic transaction tracing functionality to ArbOS. ArbOS emits a debugprint for (almost) every transaction, giving information about internal transaction calls, returns, creates, and create2s within the transaction.  
* Add a document describing how ArbOS emits this data and how it is formatted.


arb_os/evmCallStack.mini
```diff
@@ -792,9 +847,13 @@ public impure func evmCallStack_returnFromCall(
...
                 globalCurrentTxRequest.feeStats
             )?;
+
             if (completed) {
                 if (globalCurrentTxRequest.isConstructor) {
-                    // Constructor call was successful, so finish setting up the account
+                    // Constructor call was successful, so finish setting up the account.
+                    // The constructor's returndata is EVM code for the contract.
+                    // Compile that EVM code into AVM, and substitute that in as the code for the contract.
+                    let (entryPoint, evmJumpTable, _) = translateEvmCodeSegment(bytestream_new(returnData), false);
```
hkalodner, line 856: I think this is a regression introduced while merging

hkalodner (changes requested):

arb_os/tracing.mini
```diff
@@ -0,0 +1,116 @@
...
+    loop {
+        if let Some(res) = stack_pop(theStack) {
+            let (us, item) = res;
+            theStack = us;
+            val = (item, val);
+        } else {
+            asm((20000, val),) { debugprint };
+            return;
+        }
+    }
```
hkalodner, line 42: If there's a non-negligible cost to this, we could just output `currentEvmTracer.recordStack` directly

arb_os/tracing.mini
```diff
@@ -0,0 +1,116 @@
...
+}
+
+public impure func evmTracer_emitAndClear() {
```
hkalodner, line 45: I don't think this is called anywhere?

hkalodner (changes requested):

arb_os/evmCallStack.mini
edfelten: Fixed.  Good catch!

edfelten: Fixed.  Good catch.

arb_os/tracing.mini
edfelten: Fixed.  It's now called by emitTxReceipt.

arb_os/tracing.mini
edfelten: This is probably necessary, because the representation of stacks is complicated now.  (It was optimized to cram multiple stack items into a single tuple.)

hkalodner (approved):
LGTM

PR #427, opened 2021-04-02
Fix regression in error message handling

RGates94:
Fixes regression in error message handling introduced by #413 .


edfelten (approved):
Looks good.

PR #436, opened 2021-04-15
Serialize constants as json file

RGates94:
This PR causes constants to be determined from a json file, rather than being hard coded in the compiler.


arb_os/constants.json
```diff
@@ -0,0 +1,836 @@
...
+      "BLSSignatureDomainBase",
+      "73a92f91d473214defd5ffa91d036007eb2e6487fffaa551835e988fb24aaa2b"
+    ],
+    [
+      "EVMLogTopicForL2ToL1Send",
+      "5baaa87db386365b5c161be377bc3d8e317e8d98d71a3ca7ed7d555340c8f767"
+    ]
+  ],
+  "contract": [
+    "ArbAggregator",
```
hkalodner, line 831: Might be best to have these be relative paths so that the format is more extensible in the future

src/compile/miniconstants.rs
```diff
@@ -4,284 +4,85 @@
...
+use std::path::Path;
+
+#[derive(Clone, Default, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct ConstantsFile {
+    integer: Vec<(String, u64)>,
+    hex: Vec<(String, String)>,
+    contract: Vec<String>,
+}
 
 pub static ARBOS_VERSION: u64 = 5;
```
hkalodner, line 23: We'll need to get the version from the file if we need it. This constant should be removed since since it won't be synchronized with the json file

src/compile/miniconstants.rs
```diff
@@ -4,284 +4,85 @@
...
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::fs::File;
+use std::io::Read;
+use std::path::Path;
+
+#[derive(Clone, Default, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct ConstantsFile {
+    integer: Vec<(String, u64)>,
+    hex: Vec<(String, String)>,
```
hkalodner, line 19: Is there a reason for using a `Vec` instead of a `HashMap` for this?

hkalodner (changes requested):

src/compile/miniconstants.rs
RGates94: This was done just as a way of mirroring the implementation from before, I agree a map like structure would be best here. However, I would prefer a `BTreeMap` over a `HashMap` in order to get deterministic iteration.

hkalodner: That sounds even better haha. I don't know the rust standard library particularly well 😁

hkalodner (changes requested):
It seems surprising that arbos.mexe would change with this PR since the arbos code shouldn't have changed at all

edfelten (approved):
Looks good.

PR #437, opened 2021-04-15
natspec format for retryable transaction inline docs

DZGoldman:


contracts/arbos/builtin/ArbRetryableTx.sol
```diff
@@ -1,41 +1,70 @@
...
-    // Revert if called by an L2 contract, or if txId does not exist, or if txId reverts.
-    // If this returns, txId has been completed and is no longer available for redemption.
-    // If this reverts, txId is still available for redemption (until it times out or is canceled).
+
+    /**
+    * @notice Redeem a redeemable tx.
+    * Revert if called by an L2 contract, or if txId does not exist, or if txId reverts.
+    * If this returns, txId has been completed and is no longer available for redemption.
+    * If this reverts, txId is still available for redemption (until it times out or is canceled).
+    @param txId unique identifier of retryabale message: keccak256(requestID, uint(0) )
```
edfelten, line 14: typo: "retryabale"

contracts/arbos/builtin/ArbRetryableTx.sol
```diff
@@ -1,41 +1,70 @@
...
+    * @return lifetime in seconds
+    */
     function getLifetime() external view returns(uint);
 
-    // Return the timestamp when txId will age out, or zero if txId does not exist.
-    // The timestamp could be in the past, because aged-out txs might not be discarded immediately.
+    /**
+    * @notice Return the timestamp when txId will age out, or zero if txId does not exist.
+    * The timestamp could be in the past, because aged-out txs might not be discarded immediately.
+    * @param txId unique identifier of retryabale message: keccak256(requestID, uint(0) )
```
edfelten, line 27: typo: "retryabale"

edfelten: Also, "requestID" will be mysterious here.  Perhaps we shouldn't say here how the unique ID is computed, or perhaps we should explain what "requestID" is.

edfelten (changes requested):
A couple of typos, and one question about how to explain where the txId comes from.

contracts/arbos/builtin/ArbRetryableTx.sol
DZGoldman: included the full computation, bc why not

edfelten (approved):
Looks good,.

PR #439, opened 2021-04-16
Natspec arbsys

DZGoldman:


contracts/arbos/builtin/ArbSys.sol
```diff
@@ -1,31 +1,59 @@
 pragma solidity >=0.4.21 <0.7.0;
 
+/**
+* @title Precompiled contract that exists in every Arbitrum chain at address(100), 0x0000000000000000000000000000000000000064. Exposes a variety of system-level functionality.
+ */
 interface ArbSys {
-    // Get ArbOS version number
+    /**
+    * @notice  Get ArbOS version number
+    * @return version number int
```
edfelten, line 9: Maybe should say that this is the internal version number identifying an ArbOS build.

contracts/arbos/builtin/ArbSys.sol
```diff
@@ -1,31 +1,59 @@
...
+     */
     function arbOSVersion() external pure returns (uint);
 
-    // Get Arbitrum block number
+    /**
+    * @notice Get Arbitrum block number (Arbitrum genisis starts at 0, increments with qqq)
```
edfelten, line 14: When it increments is a bit complicated.  It increments whenever the Ethereum block number or timestamp of a transaction is greater than the previous transaction.  But the sequencer can also decide to cause it to be incremented even if the L1 blocknum / timestamp doesn't change.   What we can guarantee is that all txs within the same Arbitrum blocknum will have the same Ethereum blocknum and timestamp.

edfelten (changes requested):
Looks good overall.  Two comments on the content.

contracts/arbos/builtin/ArbSys.sol
DZGoldman: Yeah this will be documented elsewhere, maybe here just best to clarify that it's distinct from the L1 block number

edfelten (approved):
Looks good.

PR #442, opened 2021-04-22
Update description of retryable tickets message type 

DZGoldman:


edfelten (approved):
Thanks for fixing this.

PR #443, opened 2021-04-22
Type error reporting improvments

RGates94:
This adds code that locates where the first mismatch between a pair of types is, and uses this code to improve error message reporting.


edfelten (approved):
Looks good.

PR #444, opened 2021-04-23
Clean up no-longer-used queues

edfelten:
This fixes a storage leak in the management of the inbox's batch queue.


hkalodner (approved):
LGTM

PR #447, opened 2021-04-28
Upgrade code typechecking improvements

RGates94:
This improves typechecking for the upgrade code generator so that types match their internal representations, rather than using `any` for named types.


edfelten (changes requested):
This looks pretty good. It's a big improvement in safety from the old version.

The only issue I see is in the naming of the types representing pre-upgrade nominal types. These have names like "inbox_66".  Rather than "66" it would be much clearer to use the actual name of the type from the pre-upgrade code.  Also, the names should probably have some prefix, like maybe "preupgrade_" or "__" to avoid collision or confusion with post-upgrade types.

edfelten (approved):
Looks good.

PR #449, opened 2021-04-29
Use release mode more frequently in makefile

RGates94:
As we no longer need to rebuild the compiler each time, use of release mode doesn't incur a significant time cost and thus reduces overall time to run make.


RGates94:
This isnt necessary

PR #450, opened 2021-04-30
Change addresses of built-in contracts

edfelten:
This changes the addresses of all of Arbtrum's built-in contracts.  A contract that was previously at address `100 + x` will now be at `0xA4B0500...00 + x`. 

The address reserved for internal use by the EthBridge is now 0xA4B0500...0100

This also moves some Rust-side functionality between source files.


edfelten:
Making this a draft because we'll need to update arbos_before.mexe to get the latest version number before merging it into develop.

edfelten:
Closing this.   It was triaged pre-launch, and would have to be done differently in a post-launch chain (if it's ever needed).

PR #452, opened 2021-05-03
Add arbsys precompile function for expected return data size of current call

### LINKED ISSUE #451: Add builtin to query expected call return size
As of now you can do the low level query `returndatasize()` to get the size of the last returnData ([eip 211](https://eips.ethereum.org/EIPS/eip-211)).

It would be useful to be able to query the expect return data size for the current call.

Function selectors are the hash of the `function name with the parenthesised list of parameter types`, but this doesn't include the expected output of the function. 

If we are able to add a query for the expected output of the current function call (as supplied in the [low level call](https://github.com/OffchainLabs/arb-os/blob/e570a47764192663bd64a650f3d5f8563bf07810/arb_os/evmOps.mini#L701)) smart contracts can then differentiate certain return types by their expected length supplied by the caller.

fredlacs:
Fixes #451


edfelten:
Fixes #451

edfelten:
Closing because this was motivated by a design problem in the token bridge, but it doesn't solve that problem.

PR #455, opened 2021-05-06
Bug fix and update derived files

edfelten:
Properly initialize prices when fees are first enabled.


hkalodner (approved):
LGTM other than needing  to regenerate files again

PR #456, opened 2021-05-07
Add test that deliberately causes chain congestion

edfelten:
* Add a method to the ArbosTest built-in contract that burns a specified amount of ArbGas. If the caller asks to burn more gas than they have remaining, it burns all of their gas.
* Add a test that deliberately saturates the chain, and makes sure that a tx eventually gets rejected due to congestion.


arb_os/arbosTest.mini
```diff
@@ -39,20 +39,19 @@ use evmCallStack::evmCallFrame_getAccount;
 
 public impure func arbosTest_txcall() {
     if let Some(topFrame) = evmCallStack_topFrame() {
-        if (topFrame.caller != address(0)) {
-            evmOp_revert_knownCodePc(address(const::Address_ArbosTest), 0, 0, 0);
-        }
```
hkalodner, line 44: Don't we want to keep this protection, or are the dangerous methods protected elsewhere?

hkalodner (changes requested):

arb_os/arbosTest.mini
edfelten: They're protected elsewhere. Every dangerous method verifies that it is being called by address zero (because they're meant only for testing).

hkalodner (approved):
LGTM

PR #460, opened 2021-05-12
Rescale ArbGas by factor of 100

edfelten:
This re-scales ArbGas, so that one unit of ArbGas after the rescaling == 100 units before the rescaling. An operation that used N ArbGas before will use N/100 ArbGas after.  The supply of ArbGas will be reduced by a factor of 100, and the price will increase by a factor of 100.


hkalodner (approved):
LGTM

PR #461, opened 2021-05-12
Try to increment caller's sequence number when rejecting tx in gas auction

edfelten:
When a transaction request is rejected in the gas auction (for congestion or insufficient gas or insufficient funds for gas), this now increments the sender's sequence number (nonce), conditional on the message's sequence number having the expected value.  (Previously, the sender's seq num was unchanged in this situation.)

This should cause Metamask to get confused about sequence numbers less often.


hkalodner (approved):
LGTM

PR #463, opened 2021-05-13
Updated Makefile to yarn install packages before building

RachelFranks:
The packages in ./contracts were assumed to be already installed according to the previous build script


RGates94 (approved):
Looks good.  This will be really helpful for getting people set up with the repo.

PR #466, opened 2021-05-14
Add a fixed position global

RGates94:
This fixes `__fixedLocationGlobal` into the first global slot if it is declared in main.  There are some limitations to this solution, as the path to the global will change if 58 or more globals exist in the program, and adds the requirement that globals in main must appear first in the program as a necessary invariant.

This also changes the dispatch mechanism used to execute EVM operations as ArbOS library calls. The transpiler-generated dispatch code now reads the fixed location global to get a reference to a dispatch function, then calls that dispatch function, passing it an index that identifies which EVM operation is desired.  Because the global that points to the dispatch function is never relocated across upgrades, an upgrade can install a new dispatch function and that function will still be callable by code that was transpired before the upgrade.


edfelten:
Superseded by #487

PR #467, opened 2021-05-15
Optimize arbos code

edfelten:
Rewrite small sections of ArbOS code to speed up bottlenecks identified using the profiler.


stdlib/bytestream.mini
```diff
@@ -105,6 +105,21 @@ public func bytestream_get256(bs: ByteStream) -> option<(ByteStream, uint)> {
...
+    // read part of the next word -- just get the first nbytes bytes of it
+    // nbytes must be <= 32
+    return xif (nbytes > 32) {
+        None<(ByteStream, uint)>
+    } elseif (bs.currentOffset+nbytes > bs.capacity) {
+        None<(ByteStream, uint)>
+    } else {
+        Some((
+            bs with { currentOffset: bs.currentOffset+nbytes },
+            asm(256-8*nbytes, getbuffer256(bs.buf, bs.currentOffset+bs.slice)) uint { shr },
```
hkalodner, line 118: Would be be worth actually caching the `getbuffer256` result when we're reading one by at a time and then reading out of the cache?

arb_os/codeSegment.mini
```diff
@@ -116,15 +117,22 @@ impure func parseEvmCodeSegment(bs: ByteStream) -> EvmInstructionStack {
...
+                } else {
+                    let val = 0;
+                    while (bytesRemaining > 0) {
+                        if let Some(res) = bytestream_getByte(bs) {
+                            let (ubs, v) = res;
+                            bs = ubs;
+                            val = 256*val + v;
+                        }
+                        bytesRemaining = bytesRemaining-1;
                     }
```
hkalodner, line 133: Is is possible to hit this case?

hkalodner (changes requested):

arb_os/codeSegment.mini
edfelten: There is a corner case where that could happen, when the input byte stream ends unexpectedly before getting to the end of the expected immediate value length.  That does happen in practice, when we're parsing non-code data at the end of a file (without realizing that it's not code).

stdlib/bytestream.mini
edfelten: It's not clear whether that would be faster. That could involve more procedure calls, and has non-trivial performance tradeoffs when the reads are not 32-byte aligned.

hkalodner: Sounds good

arb_os/codeSegment.mini
hkalodner: 👍

hkalodner (approved):
LGTM

PR #469, opened 2021-05-17
Inlining controls

RGates94:
This adds basic controls for inlining of functions including function and file scope annotations and controls for default inlining behavior.


edfelten (approved):
Looks good.

PR #470, opened 2021-05-18
Fix handling of recursive calls in constructors

edfelten:
This ensures that recursive calls to a contract during its constructor are handled consistently with the EVM spec. The transpiled constructor code is run, but recursive contract calls to the constructor's address go to dummy AVM code that acts like a contract that has no code. When the constructor returns successfully, the real transpiled code for the contract is put in place, as we did before.


hkalodner (approved):
LGTM

PR #471, opened 2021-05-19
Let aggregators set their base tx fee

edfelten:
This lets aggregators set the base per-tx fee that their users are charged.  This is subject to an enforced min and max, and a default value is used for aggregators who don't set their fee.

The fees are denominated in L1 gas. ArbOS converts from L1 gas to wei using its L1 gas price estimator.

The aggregator only collects this fee if the tx's sender has registered the aggregator as its preferred aggregator. 

This PR adds affordances to get and set an aggregator's base tx fee.  Also, an aggregator address argument is added to some precompile methods, where the result now depends on which aggregator one is using.


edfelten:
Superseded by #502

PR #472, opened 2021-05-19
Made errors prettier and added control flow tooling for finding unused constructs

RachelFranks:
Changes
- Beautified compiler errors and warnings so that they show code snippets and locations of problems
- Added control flow tooling that discovers unreachable statements and unused value assignments
- This renames file_name_chart to file_info_chart which is why the mexes are different


src/compile/typecheck.rs
```diff
@@ -291,6 +292,255 @@ fn inline(
...
+    let mut alive = BTreeMap::new(); // values that are alive and need to die in this scope
+    let mut born = BTreeSet::<StringId>::new(); // values brought to life within this scope
+    let mut killed = vec![]; // values this scope has killed
+
+    // algorithm notes:
+    //   anything still alive at the end of scope is unused
+    //   scopes return everything they've killed
+    //   we make the simplifying assumption that an assignment before the first let is always used later
+    //   loops are unrolled to give assignments a second chance to be killed before leaving scope
+    //   conditionals are assumed to be
```
edfelten, line 376: Looks like this comment is unfinished.

edfelten (changes requested):
Only spotted one small issue.  I'd like to see Emily's review on this.

src/main.rs
```diff
@@ -6,10 +6,11 @@
 
 use crate::compile::CompileStruct;
 use crate::link::LinkedProgram;
-use crate::pos::try_display_location;
+//use crate::pos::try_display_location;
```
RGates94, line 9: If this use isn't needed anymore it should be deleted rather than commented out.

src/pos.rs
```diff
@@ -137,23 +138,27 @@ impl fmt::Display for Location {
...
-        file_name_chart: &BTreeMap<u64, String>,
+        file_name_chart: &BTreeMap<u64, FileInfo>,
         line_break: bool,
     ) -> String {
         format!(
             "{}{}In file: {}",
             self,
             if line_break { "\n" } else { " " },
-            file_name_chart
+            /*file_name_chart
```
RGates94, line 148: This comment block should probably be deleted.

RGates94 (changes requested):
Looks really good overall, a few small suggestions.

src/compile/typecheck.rs
RachelFranks: Good catch: I've completed the comment and added additional information for further understanding about assumptions made.

src/compile/mod.rs
```diff
@@ -841,3 +1008,27 @@ impl SourceFileMap {
...
+#[serde(transparent)]
+pub struct FileInfo {
+    pub name: String,
+    #[serde(skip)]
+    pub path: String,
+    #[serde(skip)]
+    pub contents: Vec<String>,
+}
+
+impl Debug for FileInfo {
```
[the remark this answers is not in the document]
RachelFranks: I've added the additional info save for the .contents field, which shows up as "..." now since it would otherwise require printing the entire file.

src/compile/mod.rs
```diff
@@ -369,22 +380,29 @@ pub fn compile_from_file(
...
-            None,
+            vec![],
+            false,
         ))
     }
 }
 
 ///Prints the AST nodes with indentation representing their depth, currently not used.
-fn _print_node(node: &mut TypeCheckedNode, state: &String, mut_state: &mut usize) -> bool {
+pub fn _print_node(node: &mut TypeCheckedNode, state: &String, mut_state: &mut usize) -> bool {
```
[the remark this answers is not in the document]
RachelFranks: This was a mistake, which I've now undone.

src/main.rs
RachelFranks: Deleted!

src/pos.rs
RachelFranks: Deleted!

RGates94 (approved):
Looks good.

edfelten (approved):

PR #473, opened 2021-05-19
Insufficient auto redeem funds receipt

edfelten:
When a retryable submission asks for auto-redeem but the sender doesn't have enough funds, this emits a failure receipt for the auto-redeem tx (without actually trying to run that tx), using a new error code.


hkalodner (approved):
LGTM

PR #474, opened 2021-05-20
Update DataFormat to reference the inboxseqnum bitmask

fredlacs:


edfelten (approved):
Thanks for updating this.

PR #478, opened 2021-05-21
Permission senders and support multiple fair gas price senders

edfelten:
Rather than having a single "fair gas price sender" (FGPS) this allows an arbitrary number of FGPS's.

(Background: ArbOS estimates the L1 gas price by keeping a weighted running average of the actual L1 gas prices paid by transactions submitted to the L1 Arbitrum contracts by semi-trusted parties who are called FGPS's.)

Previously there was a single FGPS address.  After this PR, ArbOS keeps a set of FGPS's, and the chain owner can add/remove addresses from the set. 

This adds methods to the ArbOwner contract to add/remove addresses from the FGPS set, query whether an address is in the FGPS set, and get a full list of addresses in the set.


edfelten:
Superseded by #495

PR #479, opened 2021-05-21
Fix unused code/variable warnings

edfelten:
Remove unused code, variables, and arguments, or prefix with underscore to tell the compiler we know it's not used. This eliminates all of the new compiler warnings triggered by the compiler improvements in PR #472 .


hkalodner (approved):
LGTM

Issue #477, opened 2021-05-20
Add safe casting


RGates94:
Currently the only option we have for changing one data type to another is `unsafecast`, and this allows conversion between arbitrary types, creating the potential for errors and crashes.

I believe we should have 3 new casting types:

1. A `cast` that ensures all possible values from the origin type can be interpreted as a valid in the destination type.  For example we would allow a cast from a `bool` to a `uint` but not from a `uint` to a `bool`.  This would always result in valid code so it would be preferable.
2. A `downcast` that allows casting to subtypes, for example from `impure func()` to `func()` or `any` to `bool`, but does not allow casts between two types that do not have a subtyping relation, so `downcast`ing from `func()` to `uint` would be forbidden.
3. A `assignablecast` that only allows casts if the origin type is assignable to the destination type.  This behavior is currently implicit in the `=` operator, and by adding this cast option we can remove this behavior, allowing for stronger type guarantees.

The primary benefits of adding these casting types are that we can maintain greater type safety in ArbOs, and making the code simpler to audit.

RGates94:
I think it might be better to allow `downcast` any time there are values in the origin type that can be cast to the destination type.  For example, something like `(bool, uint)` to `(uint, bool)`. As long as the `uint` in the second tuple field is `0` or `1`, it would produce a valid cast.  The cast from `(bool, uint)` to `(bool, bool)` would have identical requirements for safety, and would be allowed by the subtyping relation, but `(bool, uint) -> (uint, bool)` would not.  Therefore, I think this is more flexible while also maintaining the same or a higher level of safety guarantees.

PR #482, opened 2021-05-24
Safe Casting

RGates94:
Adds feature described in #477


src/compile/ast.rs
```diff
@@ -712,6 +925,32 @@ pub fn type_vectors_assignable(
...
+    tvec1: &[StructField],
+    tvec2: &[StructField],
+    type_tree: &TypeTree,
+    seen: HashSet<(Type, Type)>,
+) -> bool {
+    tvec1.len() == tvec2.len()
+        && tvec1
+            .iter()
+            .zip(tvec2)
+            .all(|(t1, t2)| t1.tipe.castable(&t2.tipe, type_tree, seen.clone()))
```
edfelten, line 951: Should we insist that the field names match here?  That seems consistent with our working definition of castable.

RachelFranks (dismissed):
Looks great to me!

edfelten (changes requested):
A couple questions about the semantics of cast operations.

src/compile/ast.rs
```diff
@@ -168,6 +168,193 @@ impl Type {
...
+                        key1.downcastable(key2, type_tree, seen.clone())
+                            && (val1.downcastable(&val2, type_tree, seen))
+                    } else {
+                        false
+                    }
+                } else {
+                    false
+                }
+            }
+            Type::Option(_) => {
```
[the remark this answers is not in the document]
RGates94: My reasoning for allowing this is that it is valid to interpret `None<A>` to `None<B>` for arbitrary types, and thus the cast my produce a valid result.

src/compile/ast.rs
```diff
@@ -168,6 +168,193 @@ impl Type {
...
+            Type::Uint | Type::Int | Type::Bytes32 => match &rhs {
+                Type::Uint | Type::Int | Type::Bytes32 => true,
+                _ => false,
+            },
+            Type::EthAddress => match &rhs {
+                Type::Uint | Type::Int | Type::Bytes32 | Type::EthAddress => true,
+                _ => false,
+            },
+            Type::Bool => match &rhs {
+                Type::Uint | Type::Int | Type::Bool | Type::Bytes32 | Type::EthAddress => true,
```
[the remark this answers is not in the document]
RGates94: My reasoning for doing this was mostly logical consistency, you probably never want to cast from a `bool` to an `address` but doing so would never result in an invalid value.

src/compile/ast.rs
RGates94: I am unclear on what exactly you mean by "our working definition of castable", if you mean our current definition of `assignability`, that is replicated by `weakcast`, if you are talking about the definition proposed in #477 , this behavior matches the proposal.

src/compile/ast.rs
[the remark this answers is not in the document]
edfelten: #477 says that downcasting should be allowed in cases where one type is a subtype of the other.  But option<A> is not a subtype of option<B>.

src/compile/ast.rs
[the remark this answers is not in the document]
edfelten: The test can't be only that the AVM representation of one is always the AVM representation of a valid value of the other one.  If that were the definition, then casting from `(uint,)` to `struct { foo: uint }` would be a valid cast.

doc/MiniLanguageTutorial.md
```diff
@@ -162,6 +162,48 @@ These rules guarantee that assignability is transitive.
...
+* `V` is one of `bytes32`, `uint`, or `int` and `S` is one of `bytes32`, `uint`, or `int`
+* `V` and `S` are tuple types with the same number of fields, and each field of `V` is castable to the corresponding field of `S`,
+* `V` and `S` are fixed-size arrays of the same size, and the field type of `V` is castable to the field type of `S`,
+* `V` and `S` are arrays, and the field type of `V` is castable to the field type of `S`,
+* `V` and `S` are structs, with the same number of fields, and each field of `V` is castable to the corresponding field of `S`,
+* `V` and `S` are function types, with the same number of arguments, and either `S` is impure or `V` is not impure, and each argument type of `V` is castable to the corresponding argument type of `S`, and either (a) both `S` and `V` return void, or (b) the return type of `S` is castable to the return type of `V`.  (Note that the return type is compared for castability "backwards". This is needed to make calls through function references type-safe.)
+* `V` and `S` are map types, and the key type of `V` is castable to the key type of `S`, and the value types of `V` and `S` are castable.
```
edfelten, line 179: This should say what relationship the value types of a map need to have.  (Castability is directional. Does this mean value type of V is cartable to value type of S, or vice versa?)

doc/MiniLanguageTutorial.md
```diff
@@ -162,6 +162,48 @@ These rules guarantee that assignability is transitive.
 
 The compiler uses often uses type inference to infer the types of variables from the types of values assigned to them.  If a programmer wants the compiler to infer a different type, they should use an explicit type-casting operation to convert the value to the desired type.
 
+## Castability
```
edfelten, line 165: It would be useful to have a rationale for why cast ability is defined this way.  In particular, why does cast ability differ from assignability?  There's an implicit notion of subtyping here, so why have two different notions?

doc/MiniLanguageTutorial.md
```diff
@@ -162,6 +162,48 @@ These rules guarantee that assignability is transitive.
...
+* `V` and `S` are tuple types with the same number of fields, and each field of `V` is castable to the corresponding field of `S`,
+* `V` and `S` are fixed-size arrays of the same size, and the field type of `V` is castable to the field type of `S`,
+* `V` and `S` are arrays, and the field type of `V` is castable to the field type of `S`,
+* `V` and `S` are structs, with the same number of fields, and each field of `V` is castable to the corresponding field of `S`,
+* `V` and `S` are function types, with the same number of arguments, and either `S` is impure or `V` is not impure, and each argument type of `V` is castable to the corresponding argument type of `S`, and either (a) both `S` and `V` return void, or (b) the return type of `S` is castable to the return type of `V`.  (Note that the return type is compared for castability "backwards". This is needed to make calls through function references type-safe.)
+* `V` and `S` are map types, and the key type of `V` is castable to the key type of `S`, and the value types of `V` and `S` are castable.
+* `V` and `S` are optional types, and the inner type of `V` is castable to the inner type of `S`
+
+## Covariant Cast
```
edfelten, line 182: It would be useful to have an explanation of why this notion of covariant casting is useful.  Why not just use unsafe cast for these?  What additional safety does covariantcast provide in practice?

doc/MiniLanguageTutorial.md
```diff
@@ -162,6 +162,48 @@ These rules guarantee that assignability is transitive.
...
+* `S` is `any`,
+* `V` equals `S`,
+* both `V` and `S` are one of `bool`, `address`, `bytes32`, `uint`, or `int`
+* `V` and `S` are tuple types with the same number of fields, and each field of `V` is covariant castable to the corresponding field of `S`,
+* `V` and `S` are fixed-size arrays of the same size, and the field type of `V` is covariant castable to the field type of `S`,
+* `V` and `S` are arrays, and the field type of `V` is covariant castable to the field type of `S`,
+* `V` and `S` are structs, with the same number of fields, and each field of `V` is covariant castable to the corresponding field of `S`,
+* `V` and `S` are function types, with the same number of arguments, and each argument type of `V` is covariant castable to the corresponding argument type of `S`, and either (a) both `S` and `V` return void, or (b) the return type of `S` is covariant castable to the return type of `V`.
+* `V` and `S` are map types, and the key type of `V` is covariant castable to the key type of `S`, and the value types of `V` and `S` are covariant castable.
+* `V` and `S` are optional types
```
edfelten, line 205: This is inconsistent with the definition above.  If V is `option<uint>` and S is `option<func()>`, there is no constructible type that is castable to both V and S.

edfelten (changes requested):

doc/MiniLanguageTutorial.md
RGates94: The type `option<every>` is constructable and can be assigned to both.

doc/MiniLanguageTutorial.md
RGates94: This prevents a large number of type pairs from casting, for example you cannot cast between structs of different lengths, tuples of different lengths, arrays of different lengths, types that fall into categories, and types whose subtypes aren't covariant castable.

doc/MiniLanguageTutorial.md
RGates94: We may wish to convert from for example `bool` to `uint`, where there is not a subtyping relation.

edfelten (approved):

PR #483, opened 2021-05-24
Make EVM GAS instruction return ArbGas left, not AVM gas

### LINKED ISSUE #481: EVM GAS opcode should return ArbGas left
Currently the EVM GAS opcode returns the amount of AVM gas left. It should return the amount of ArbGas (which is smaller).

edfelten:
Fixes #481


hkalodner (approved):
LGTM

PR #484, opened 2021-05-25
ArbSys::isTopLevelCall returns true at top level of retryable

edfelten:
This make ArbSys::isTopLevelCall return true when called at the top level of a retryable tx. (Previously it returned false because the retry is wrapped inside an outer redeem tx.).

This is useful for token bridge functionality.


hkalodner (approved):
LGTM

PR #485, opened 2021-05-25
Refactor to use ETH-conserving primitives

edfelten:
This minimizes the use of ETH credit and debit operations, replacing them with ETH transfer operations which inherently hold the total ETH balances in the system constant. 

To support this, we introduce a new notion of "escrow boxes" which can hold ETH for administrative purposes. For example, when a tx starts running, the ETH that will pay for its gas is transferred into an escrow box, and those ETH will be transferred out when the tx completes.

The obvious exceptions are deposits from L1 and withdrawals to L1, which inherently need to use credit/debit.


arb_os/constants.json
```diff
@@ -107,6 +107,7 @@
     "EVMCallType_staticcall": 3,
     "EVMNonZeroBalanceCallStipend": 2300,
     "EVMWriteL1GasCost": 5000,
+    "EscrowStoreID_txGas": 0,
```
hkalodner, line 110: It looks like EscrowStoreID_txGas is set to two different values in this file

edfelten: Removed one

arb_os/gasAccounting.mini
```diff
@@ -39,6 +40,7 @@ use messages::TxRequestData;
 use std::bytearray::bytearray_new;
 use std::bytearray::bytearray_getByte;
 use std::bytearray::bytearray_get256;
+use std::bytearray::bytearray_get256;
```
hkalodner, line 43: Duplicate import

arb_os/gasAccounting.mini
```diff
@@ -584,20 +566,22 @@ public impure func gasAccounting_endTxCharges(
 
     let refund = (txState.originalMax - totalGasUsed) * txState.price;
     if (refund > 0) {
-        payments = deferredPayments_add(payments, txState.refundTo, refund);
+        payments = deferredPayments_add(payments, const::EscrowStoreID_txGas, txState.refundTo, refund);
```
hkalodner, line 569: Just confirming that the deferred payment system is still needed independently from the escrow system?

arb_os/gasAccounting.mini
```diff
@@ -1251,21 +1174,21 @@ public func deferredPayments_new() -> DeferredPayments {
     return stack_new();
 }
 
-public func deferredPayments_add(dp: DeferredPayments, recipient: address, amount: uint) -> DeferredPayments {
-    return stack_push(dp, (recipient, amount));
+public func deferredPayments_add(dp: DeferredPayments, escrowKey: uint, recipient: address, amount: uint) -> DeferredPayments {
```
hkalodner, line 1177: Looks like this is only ever called with `EscrowStoreID_txGas` as the `escrowKey`. If that's the only way this will be used, might be better to hardcode that here rather than passing it in at all the callsites

arb_os/gasAccounting.mini
```diff
@@ -1251,21 +1174,21 @@ public func deferredPayments_new() -> DeferredPayments {
...
-            acctStore = accountStore_set(
-                acctStore,
-                recipient,
-                account_addToEthBalance(accountStore_get(acctStore, recipient), amount),
-            );
+            let (escrowKey, recipient, amount) = unsafecast<(uint, address, uint)>(item);
+            if let Some(uAcctStore) = accountStore_payFromEscrow(acctStore, escrowKey, recipient, amount) {
+                acctStore = uAcctStore;
+            } else {
+                asm(11000,) { debugprint };
```
hkalodner, line 1190: Is it possible this will hit and if so would it make sense to revert?

arb_os/retrybuffer.mini
```diff
@@ -239,14 +245,20 @@ public impure func retryBuffer_delete(rbuf: RetryBuffer, txId: uint, withRefund:
...
+                    getGlobalAccountStore(),
+                    rtx.callvalueEscrowKey,
+                    rtx.beneficiary,
+                    rtx.callvalue
+                ) {
+                    setGlobalAccountStore(astore);
+                }  // else something went badly wrong and we can't refund the callvalue
+            } else {
+                // try to refund from the escrow, ignoring errors (which should never happen)
+                let _ = evmCallStack_transferEthFromEscrow(rtx.callvalueEscrowKey, rtx.beneficiary, rtx.callvalue);
```
hkalodner, line 260: Maybe good to put a debugprint for debugging in case we have an error

hkalodner (changes requested):

arb_os/gasAccounting.mini
edfelten: Removed one

arb_os/gasAccounting.mini
edfelten: Yes, deferred payments are needed because the payments have to be made at a different stack frame level.

arb_os/gasAccounting.mini
edfelten: I'd prefer to keep this as is, so it's easier to re-use the deferred payments mechanism later for other purposes.

arb_os/gasAccounting.mini
edfelten: Changed this to have the function return an option, then dealt with the failure case appropriately at each callsite.  (For all of the current callsites, ignoring the failure is the best approach.)

arb_os/retrybuffer.mini
edfelten: Added a debugprint

hkalodner (approved):
LGTM

PR #488, opened 2021-05-25
Fix returndata bug

edfelten:
Properly set up returndata and status when making an attempted call to a non-contract.


hkalodner (approved):
LGTM

PR #489, opened 2021-05-25
Charge network fees even if no aggregator

edfelten:
Charge network fees (but not aggregator reimbursement) for txs that don't have an aggregator.


hkalodner (approved):
LGTM

PR #492, opened 2021-05-26
Warnings for unused use-statements

RachelFranks:
This branch makes the following changes
- Removes all the unused use-statements in ArbOS
- Updated the grammar to allow constant declarations before use-statements
- Adds a warning system for collecting warnings throughout compilation
- Adds a -w flag (and --warnings-are-errors) that promotes warnings to errors
- Adds a test case for the warning system


RachelFranks:
Just merged the following additional features

- Creates and walks callgraph to prune and warn on unused functions
- Makes typechecking multithreaded
- Adds test case for callgraph traversal
- Improves typechecking errors and adds various file locations
- Compile errors on top-level symbol redefinition
- Compile errors on empty functions
- Builtins no longer import themselves
- Improves variable naming around paths and stringtable ids
- Adds additional metadata to imports

RachelFranks:
New: Detect unused constants and mark scopes for codegen printing

- Detects constants that are never used (includes test case)
- Added -m flag, short for --must_use_global_consts, for ArbOS
- Marked constants in ArbOS's constants.json that we intentionally don't use
- Removed "constval()" from the grammar
- Added a "print" attribute, which prints the instructions an ast node generates
- Replaced the warning system with a full error system (includes test case)
- Not all errors halt compilation => now multiple are printed at the end
- Added file locations to parser/lexer errors and warnings

RGates94 (approved):
Everything looks good to me.

PR #495, opened 2021-05-27
Permission senders and support multiple fair gas price senders for launch

edfelten:
This creates a permissions list of senders. Tx's sent by parties not on the list are rejected with a "sender not permitted" error code.

This also replaces the single "fair gas price sender" (FGPS) mechanism with a list that can contain an arbitrary number of FGPS's.

(Background: ArbOS estimates the L1 gas price by keeping a weighted running average of the actual L1 gas prices paid by transactions submitted to the L1 Arbitrum contracts by semi-trusted parties who are called FGPS's.)

Previously there was a single FGPS address. After this PR, ArbOS keeps a set of FGPS's, and the chain owner can add/remove addresses from the set.

This adds methods to the ArbOwner contract to add/remove addresses from the FGPS set, query whether an address is in the FGPS set, and get a full list of addresses in the set.


hkalodner (approved):
LGTM

PR #497, opened 2021-05-27
Check previous upgrade hash on upgrade

edfelten:
Add an argument to ArbOwner call that completes an upgrade, to pass in the hash of the previous upgrade code.  The upgrade is aborted unless either (a) argument is zero, or (b) the argument equals the hash of the previous upgrade code.

Also hard-code the chain ID for mainnet release.


hkalodner (approved):
LGTM

PR #501, opened 2021-05-29
Accept non-gas-estimation sideloads from any address

edfelten:
Accept a sideloaded message from any sender, even if the sender is not on the permitted-senders list, unless the sideload is a gas estimation message.  (Gas estimation is treated differently because the goal is to behave like the same tx would behave if it were really submitted -- so we should reject based on address if that would happen for the same tx in the non-sideload case.)


hkalodner (approved):
LGTM

PR #502, opened 2021-05-30
Aggregator controlled base price

edfelten:
Lets each aggregator control the base fee that is charged to tx's that it submits (assuming it is the tx sender's preferred aggregator).

Adds ArbAggregator methods to control this.  Aggregators who don't set their base fee will get a default fee.  A minimum and maximum on the fee are enforced.


contracts/arbos/builtin/ArbGasInfo.sol
```diff
@@ -10,10 +10,10 @@ interface ArbGasInfo {
     //            per ArbGas congestion,
     //            per ArbGas total
     //        )
-    function getPricesInWei() external view returns (uint, uint, uint, uint, uint, uint);
+    function getPricesInWei(address aggregator) external view returns (uint, uint, uint, uint, uint, uint);
 
     // return prices in ArbGas (per L2 tx, per L1 calldata unit, per storage allocation)
-    function getPricesInArbGas() external view returns (uint, uint, uint);
+    function getPricesInArbGas(address aggregator) external view returns (uint, uint, uint);
```
hkalodner, line 16: We should discuss why types of API consistency guarentees we should provide. Theoretically this change could break deployed contracts though it seems unlikely

contracts/arbos/builtin/ArbAggregator.sol
```diff
@@ -25,5 +25,13 @@ interface ArbAggregator {
...
     //      or by the chain owner.
     function setFeeCollector(address aggregator, address newFeeCollector) external;
+
+    // Get the tx base fee (in approximate L1 gas) for aggregator
+    function getTxBaseFee(address aggregator) external view returns (uint);
+
+    // Set the tx base fee (in approximate L1 gas) for aggregator
+    // If feeInL1Gas is outside the chain's allowed bounds, fee will be "clipped" to the min or max allowed value
+    // Reverts unless called by aggregator or the chain owner
+    function setTxBaseFee(address aggregator, uint feeInL1Gas) external;
```
hkalodner, line 35: Personally I would prefer reverting on bad inputs (out of allowed range) rather than silently adjusting. Would probably be a good question to raise with the group

arb_os/messageBatch.mini
```diff
@@ -121,9 +120,6 @@ public func messageBatch_get(batch: MessageBatch) -> option<(IncomingRequest, Me
             }
         ))
     } else {
-        if (batch.numSoFar > 0) {
-            updateBatchSizeEstimate(batch.numSoFar, batch.template.sender);
```
hkalodner, line 125: The definition of `updateBatchSizeEstimate` should probably be removed

arb_os/gasAccounting.mini
```diff
@@ -161,12 +158,8 @@ public impure func gasAccounting_init() {    // initialize on boot, before we ha
     setGasAccountingParamsFromSpeedLimit(const::DefaultSpeedLimitPerSecond);
 
     let pricingParams = struct {
-        l1GasPerL2Tx: const::Charging_DefaultL1GasPerL2Tx,
-        arbGasPerL2Tx: const::Charging_DefaultArbGasPerL2Tx,
         l1GasPerL1CalldataUnit: const::Charging_DefaultL1GasPerCalldataUnit,
-        arbGasPerL1CalldataUnit: const::Charging_DefaultArbGasPerCalldataUnit,
         l1GasPerStorage: const::Charging_DefaultL1GasPerStorage,
-        arbGasPerStorage: const::Charging_DefaultArbGasPerStorage,
```
hkalodner, line 169: Looks like some/all of these constants may be unused after this. Would be good to remove them, and also to have @RachelFranks's PR that would warn on these merged

arb_os/gasAccounting.mini
```diff
@@ -371,24 +395,20 @@ public impure func gasAccounting_setFeesEnabled(enabled: bool) {
 }
 
 public impure func gasAccounting_setChargingParams(
-    l1GasPerL2Tx: uint,
-    arbGasPerL2Tx: uint,
+    _l1GasPerL2Tx: uint,
+    _arbGasPerL2Tx: uint,
```
hkalodner, line 399: Shouldn't the unused args be removed?

arb_os/gasAccounting.mini
```diff
@@ -1181,10 +1210,20 @@ public func txFeeStats_zero() -> TxFeeStats {
     };
 }
 
-public func txFeeStats_setPrices(stats: TxFeeStats, prices: GasPrices) -> TxFeeStats {
+public impure func txFeeStats_setPrices(stats: TxFeeStats, prices: GasPrices, aggregator: option<address>) -> TxFeeStats {
```
hkalodner, line 1213: Unclear why this became impure?

hkalodner (changes requested):

contracts/arbos/builtin/ArbGasInfo.sol
edfelten: I'm going to change this to leave the old no-argument method in place and add a new method that takes an aggregator address argument.  The no-arg method will call the new method, passing the address of the sender's preferred aggregator.

contracts/arbos/builtin/ArbAggregator.sol
edfelten: Added an argument to let the caller choose which behavior they want on out-of-bound value: revert or clip.

arb_os/messageBatch.mini
edfelten: Done.

arb_os/gasAccounting.mini
edfelten: Removed the unused constants.

arb_os/gasAccounting.mini
edfelten: Done.

arb_os/gasAccounting.mini
edfelten: It's impure because it looks up the tx base fee for an aggregator address, which is stored in that aggregator's account in the global account table.

arb_os/gasAccounting.mini
```diff
@@ -886,111 +907,160 @@ public impure func getNextRequestFromCongestionAuction() -> (
...
+                } with {
+                    calldataUnit: feeStats.weiPaid.calldataUnit + weiForCalldata
+                }
+            },
+            totalDue,
+        )
+    } else {
+        // this should never happen, because we reduced totalDue to fit within payer's funds
+        // so just do something safe
+        (
```
PlasmaPower, line 1058: Could we add a healthcheck or at least debugprint here?

PlasmaPower (dismissed):
LGTM

arb_os/gasAccounting.mini
edfelten: Added a debugprint.

PR #505, opened 2021-05-31
Let owner designate a trusted L1 gas price oracle

edfelten:
This creates a chain parameter which designates an address as a trusted oracle for the L1 gas price. The oracle (if any) or chain owner can call a new ArbGasInfo method to set ArbOS's L1 gas price estimate. 

This also adds a getter for the L1 gas price estimate.

This deprecates the L1 gas price setter method in ArbOwner, in favor of the new one in ArbGasInfo.


rachel-bousfield:
This has been merged :)

PR #508, opened 2021-05-31
Update arbos_before.mexe after mainnet upgrade 1

edfelten:
This updates the before version of ArbOS, and the upgrade tests, after mainnet upgrade 1 (earlier today). 

Also removes some tests that no longer make sense.


hkalodner (approved):
LGTM

PR #509, opened 2021-06-02
Panic on excess globals

RGates94:
Because of the behavior of `__fixedLocationGlobal`, ArbOs must have 57 or fewer global variables. This PR causes the compiler to panic if this invariant is not met.


edfelten (approved):
Looks good.

PR #513, opened 2021-06-05
Refactor inbox message handling

edfelten:
This refactors the handling of arriving inbox messages, moving toward a "push" approach rather than "pull".  This means that the inbox code calls into downstream components when it has a request to process, rather than those components calling back into the inbox code to get the next request.  This makes the code simpler and easier to understand.

This also leads to some useful refactoring of parts of the gas accounting code.


rachel-bousfield:
We've merged this already :)

PR #514, opened 2021-06-07
Compiler optimization passes and inlining

rachel-bousfield:
Optimizations, inlining, and tooling

- Optimization passes that reduce ArbOS gas costs 8-12%
- New codegen-level inlining for splicing in stack frames
- Control Flow Graphs for the current IR
- Optimization tooling for comparative analysis and bug discovery
- Added -s flog, short for --show_optimizations, to pretty-print details
- Made minor simplification to function codegen
- Improved attribute propagation to cover asm() and structs
- Corrected child_nodes() bug around struct field assignments


rachel-bousfield:
While the passes are slightly different, the main optimization strategies in this PR constitute
- #631

PR #515, opened 2021-06-07
Mini libraries and ArbOS code coverage

edfelten:
Adds a code coverage feature to the test suite.  Running `make coverage` will generate a set of coverage files for all of the non-ArbOS tests, along with an `alltests.cov` file that is the union of the coverage sets of all of the individual tests.


RachelFranks:
Merged in the following language tweaks
- ASM commas are now optional
- Multi-line comments are now supported

PR #516, opened 2021-06-10
Assert Statements

RachelFranks:
- Adds a typechecked "assert" builtin and ast node
- Completes attribute propagation


src/mini.lalrpop
```diff
@@ -118,9 +118,10 @@ StatementKind: StatementKind = {
 	"return" <e: Expr> ";" => StatementKind::Return(e),
 	"return" ";" => StatementKind::ReturnVoid(),
 	<lno: @L> "return" "None" ";" => StatementKind::Return(Expr{ kind: ExprKind::Constant(Constant::Option(OptionConst::None(Type::Every))), debug_info: DebugInfo::from(file_info.location(BytePos::from(lno), filename))}),
-	"break" <e: Expr?> ";" => StatementKind::Break(e,None),
-	"debug" "(" <e: Expr> ")" ";" => StatementKind::DebugPrint(e),
-	"asm" "(" <a:CommaedExprs?> ")" "{" <body:AsmInsn*> "}" ";" => StatementKind::Asm(body, a.unwrap_or(vec![])),
+	"break" <e: Expr?> ";" => StatementKind::Break(e, None),
+        "debug" "(" <e: Expr> ")" ";" => StatementKind::DebugPrint(e),
+        "assert" "(" <e: Expr> "," <p: Expr> ")" ";" => StatementKind::Assert(e, p),
```
edfelten, line 123: Let's discuss the syntax we want for this.  It seems to me that there might be a bunch of overlap between the code that decides whether the assertion passes, and the code that generates the message to print if it doesn't pass.  This syntax would require the programmer to duplicate code in that case.

I can think of two alternatives to discuss.  One would be to write `assert( [expr] )` where expr should have type `(bool, any)`.  Another is to use the single-expo syntax but with the expression having type `option<any>` where `None<any>` means the assertion is satisfied, and `Some(messageToPrint)` means it is not and `messageToPrint` should be printed.

edfelten (commented):
Code looks fine.  I'd like to discuss what the syntax should be for this.

src/mini.lalrpop
RachelFranks: Code duplication definitely makes the current syntax sub-optimal. My preference is that we use options, but we could even do both in cases where returning a None feels awkward. I think either is great.

edfelten (approved):
Looks good.

RGates94 (approved):
Looks good

PR #519, opened 2021-06-15
Untagged union

RGates94:
This adds untagged unions to mini.  

They can be created by `newunion<type1, type2>(expr)`, and can be cast to their component types by `unioncast<component>(expr)`.


edfelten (changes requested):
I think we need to add syntax to declare that something (global variable, function arg, struct field, etc.) has a union type.  I don't see anything in added to the grammar to support that.

edfelten (dismissed):
Looks good.

doc/MiniLanguageTutorial.md
```diff
@@ -102,6 +102,10 @@ Mini has the following types:
 
 > a hash map, which maps keys of one type to values of another type
 
+`union<` *type1*, *type2*, ... `>`
+
+> an untagged union type that can values from any of *type1*, *type2*, etc. There must be explicit casts to convert to and from its component types. This is done via `newunion` and `unioncast`.
```
edfelten, line 107: typo here "that can values"

edfelten (changes requested):
One typo in the language spec.

edfelten (approved):
Thanks for fixing the typo.

PR #520, opened 2021-06-15
Shifts like <<

RGates94:
This changes the syntax for left and right shift to use `<<` and `>>` as is common in other languages.  

This breaks backward compatibility in situations where two closing brackets are used in sequence, for example `unsafecast<option<uint>>` will now be interpreted as an error due to the `>>` being interpreted as a right shift.  This can always be fixed by adding whitespace between the `>`s, for example `unsafecast<option<uint> >`.

All necessary fixes to the existing mini code and code generator have already been applied.


edfelten (changes requested):
This should also replace all of the `asm` expressions that use `shl` and `shr` with the corresponding `<<` and `>>` expressions.

Otherwise, this looks good.

src/uint256.rs
```diff
@@ -410,6 +410,10 @@ impl Uint256 {
         let hash_result = keccak256(&bytes1);
         Uint256::from_bytes(&hash_result)
     }
+
+    pub fn non_negative(&self) -> bool {
```
edfelten, line 414: I think this will always return true.  (uint256's are non-negative by definition -- hence the name).

edfelten (changes requested):
One bug, and a couple of type-checking policies to discuss.

edfelten (approved):

PR #523, opened 2021-06-15
Return different errors on seqnum too low vs too high

### LINKED ISSUE #522: Give different error codes if sequence number is too large vs too small
Some applications want to be able to distinguish these cases.

edfelten:
This adds new tx error codes 14 (sequence number / nonce too low) and 15 (sequence number / nonce too high).  ArbOS returns each one when appropriate.

Tx error code 5 (bad sequence number / nonce) is no longer used.

Fixes #522


hkalodner (approved):
LGTM

PR #525, opened 2021-06-18
Add special-case upgrade code for WETH9 contract

edfelten:
Add special-case code to the ArbOS upgrade path, to patch an issue with one of the WETH9 contracts. This modifies one location in the contract's storage.


hkalodner (approved):
LGTM

PR #527, opened 2021-06-21
Mixed let

### LINKED ISSUE #491: Add mixed let statement
It would be beneficial to have a mixed let statement, that assigns some fields of a match pattern to new variables, and others to existing variables.  One possible syntax for this would be:

```
func main() {
    let y = 4;
    (let x, y) = f(y);
}

func f(y: uint) {
    return (y, y+1);
}
```
This would allow the use of fewer local variables for the same logic, saving gas.

RGates94:
This adds a mixed let, which allows assignments and new bindings in the same tuple let statement.  Resolves #491 .


edfelten (changes requested):
Please revise the language spec to describe the correct syntax and functionality.

doc/MiniLanguageTutorial.md
```diff
@@ -212,9 +212,17 @@ Values of type `anytype` do not have any representation that is understood by th
...
-> Create multiple new variables based on unpacking a tuple. *expression* must be a tuple type, with the number of fields in the tuple equal to the number of names on the left-hand side.  The compiler creates a new local variable for each name on the left-hand side, and infers the type of each new variable based on the type of the corresponding field of the right-hand side tuple.
+> Where each nameorbinding may be:
+
+*identifier*
+
+> Or
+
+**identifier*
+
+> Creates or assigns to multiple variables based on unpacking a tuple. If the *nameorbinding* is an identifier it creates a new variable, if *nameorbinding* is **identifier* it assigns to an existing variable with that name.  *expression* must be a tuple type, with the number of fields in the tuple equal to the number of names on the left-hand side.  The compiler creates a new local variable for each name on the left-hand side, and infers the type of each new variable based on the type of the corresponding field of the right-hand side tuple.
```
edfelten, line 225: I think this should say it creates a new variable for each non-starred name on the left-hand side.

src/compile/ast.rs
```diff
@@ -1016,6 +1016,7 @@ pub struct MatchPattern<T = ()> {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub enum MatchPatternKind<T> {
     Simple(StringId),
+    Assign(StringId),
```
edfelten, line 1019: Is there a better name for the non-assignment one than `Simple`?  That made sense back when it was the only non-tuple kind, but another name might be clearer now.

edfelten (changes requested):
A couple of small issues to consider.

src/compile/ast.rs
RGates94: I have renamed this to `Bind`.

RachelFranks (changes requested):
The following test program induces an out-of-bounds local-variable write
```
func main() -> uint {
    let x = 8;
    let (*x, y) = (8, 8);
    return x;
}
```
```
Compile error: TupleTree::write_code: Linking error: out-of-bounds write
    --> minitests/mixed.mini line 5 column 5
     |
5    |     let (*x, y) = (8, 8);
     |     ^
```

RachelFranks:
I've made the requisite changes to make liveliness analysis work again. This way warnings should be complete and accurate when using this feature in ArbOS.

- Extends liveliness flow checker to work for mixed let
- Warnings now line the arrow up correctly for simple lets

RachelFranks (approved):

edfelten (approved):

PR #528, opened 2021-06-22
Only do WETH fix on mainnet

edfelten:
On upgrade, this checks the chain ID and only does the WETH contract fix if it's running on mainnet.


hkalodner (approved):
LGTM

PR #531, opened 2021-06-23
Gasprice zero means zero

### LINKED ISSUE #529: Remove special treatment of gasPrice==0
When a user tx sends a gasPrice of zero, ArbOS special-cases the zero and treats it as a request to pay the current ArbGas price, whatever it is.  This special-case behavior is no longer needed, and confuses some tooling, so we should discontinue it.

edfelten:
In submitted transactions, if the submitted gas price is zero, treat it as zero. 

Previously, we special-cased a gas price of zero, interpreting it as "charge me whatever the current gas price is".  This was convenient for testing, but it would confuse some tooling. 

Closes #529


RachelFranks:
This was merged into our currently deployed code

PR #532, opened 2021-06-23
Remove pluggables

edfelten:
This removes the pluggable modules subsystem from ArbOS.  It's not needed now, and given the improvements in upgrade support, it's probably not very useful going forward.


rachel-bousfield:
This has been integrated and deployed

PR #533, opened 2021-06-23
No return address functions

RGates94:
This adds the ability to create functions of the form:

```
func main() {
    let q = 1;
    let x = test();
}

func test() -> every {
    return xloop {
        let x = 4;
        let z = 2;
    };
}
```

Where the compiler does not generate a return address for `test()`, and where `x` is of type `every`.  As `every` can be assigned to arbitrary types, these functions can be used in a wide variety of locations , similar to how `panic!()` or `unimplemented!()` work in rust.


src/compile/mod.rs
```diff
@@ -54,6 +54,8 @@ pub struct CompileStruct {
     pub consts_file: Option<String>,
     #[clap(short, long)]
     pub release_build: bool,
+    #[clap(short, long)]
+    pub no_builtins: bool,
```
edfelten, line 58: Can the compiler figure out whether to link in the builtins, and just include the builtins that are actually used by the program?  Seems like this could be better than making the programmer guess.

src/mini.lalrpop
```diff
@@ -159,6 +159,7 @@ Type: Type = {
 	<i: "impure"?> "func" <a: CommaedTypesInParens> <r: ("->" <Type>)?> => Type::Func(i.is_some(), a, Box::new(r.unwrap_or(Type::Void))),
 	"any" => Type::Any,
 	"option<" <t1:Type> ">" => Type::Option(Box::new(t1)),
+	"every" => Type::Every,
```
edfelten, line 162: The syntax here will be mysterious to many people who are reading Mini code.  It's true in a formal sense that a program that returns type `every` cannot have any returns, but that requires understanding what the type `every` (which never occurs in a program) means, what its implication is, and that you declare a return type in order to declare that the function never returns.  

Can we add a more straightforward syntax as an alternative?  That could just be syntactic sugar for the `-> every` construct.

edfelten (commented):
A couple of issues to discuss.

src/compile/mod.rs
RGates94: This would be significantly more work, and the immediate benefit of even the manual option is large, so I do not believe it would be beneficial to add automatic functionality to this PR.

src/mini.lalrpop
```diff
@@ -61,10 +61,15 @@ TypeDecl: TypeDecl = {
...
 
 FuncDecl: Func = {
-	<lno: @L> <attribs: ("#[" <Attributes> "]")?> <p: "public"?> <imp: "impure"?> "func" <i:Ident> "(" <fa:FuncArgs> ")" <t: ("->" <Type>)?> <cb:CodeBlock> =>
+	<lno: @L> <attribs: ("#[" <Attributes> "]")?> <p: "public"?> <imp: "impure"?> "func" <i:Ident> "(" <fa:FuncArgs> ")" <t: (Return)?> <cb:CodeBlock> =>
 	        Func::new(i, imp.is_some(), fa, t.unwrap_or(Type::Void), cb, p.is_some(), DebugInfo::new(file_info.location(BytePos::from(lno), filename), attribs.unwrap_or_default())),
 }
 
+Return: Type = {
+    "->" <Type> => <>,
+    "noreturn" => Type::Every,
```
edfelten, line 70: Can you add this to the spec too?   After that, we'll be good to go.

edfelten (changes requested):
One small spec change requested.

edfelten (approved):

PR #534, opened 2021-06-24
SetGas, PushGas, and asm renaming

RachelFranks:
Made instruction naming consistent with the Arbitrum design spec.
Removed SetGas and PushGas asm in favor of new expressions.


edfelten (approved):
Looks good.  I'm approving this, but let's talk about how and when to merge it.

edfelten (changes requested):
I think we need to make the setgas and gets language-level functions impure.

RachelFranks:
> I think we need to make the setgas and gets language-level functions impure.

Just made the changes: SetGas(_) and GetGas() make a function impure.

edfelten (approved):
Looks good.

PR #535, opened 2021-06-24
Add getter and check for last ArbOS upgrade hash

### LINKED ISSUE #530: In ArbOwner, add getter for previousArbosUpgradeHash
Adding a getter for this would make upgrading easier. There's not a safety issue, because this is checked on every upgrade, but currently we pay for uploading the upgrade before we check.   Better to have a getter so we can check offline before doing the upload.

edfelten:
Change ArbOwner to
* add a getter that returns the code hash of the last ArbOS upgrade, or zero if there hasn't been an upgrade on this chain
* add a `lastArbosUpgradeHash` to `startCodeUpload`, which will cause that method to revert unless the provided value is either zero or is equal to the code hash of the last ArbOS upgrade. This lets the owner verify the previous code hash before starting to upload an upgrade.


rachel-bousfield:
This has already been merged :)

PR #536, opened 2021-06-24
Fix ecmul in the emulator to handle invalid slices found from fuzzing

RachelFranks:


edfelten (approved):
Looks good.  First find from the fuzzing!

PR #539, opened 2021-06-29
Beautify with-statement syntax

RachelFranks:
- adds support for left-hand-side struct, map, and array assignments

```
// new syntax
set blockhashes.oldHashes[b % 256] = hash(bytes32(b), inboxAccumHash);

// old syntax
blockhashes = blockhashes with {
    oldHashes: blockhashes.oldHashes with { [b % 256] = hash(bytes32(b), inboxAccumHash) }
};

// ready for arbitrary nesting (with doesn't currently support this)
set this.that.those[8].them = prices.perTx;
```


edfelten (approved):
Looks good.

PR #541, opened 2021-07-01
Fix buffer equality

rachel-bousfield:
The derived PartialEq intractably goes through all nodes,
so we now just check that the hashes are equal.


rachel-bousfield:
This was equivalently introduced in 
- #644

PR #549, opened 2021-07-10
Check nonce on create opcodes

### LINKED ISSUE #548: CREATE2 should fail if target address nonce is nonzero
EIP-1014 says that the EVM instructions CREATE2 should fail (not create a contract, and return zero) if the address where the new contract would be deployed has nonzero nonce.  Currently this operation succeeds.  

The same issue holds for CREATE.

This should be changed to match the specified behavior.

edfelten:
When emulating the CREATE and CREATE2 opcodes, this checks that the address where the new contract would be deployed has nonce zero, and fails the operation (not creating a contract, and returning address zero) otherwise.

Fixes #548


rachel-bousfield:
Already merged!

PR #552, opened 2021-07-11
Gas control tweaks

### LINKED ISSUE #542: Limit max increase in retryable cost
The cost of submitting a retryable is updated once per day (on mainnet) at a predictable time.  This poses UX issues around the time of the update, because users can't easily predict how much their submission will cost (in the case where the submission lands on L2 just after the update).

To mitigate this, we should put an upper bound on the increase, so that a dev can ensure that their submission won't get rejected.

### LINKED ISSUE #543: Auto-redeem doesn't work with fees disabled
When fees are disabled, ArbOS doesn't try to auto-redeem submitted retryables.  Presumably this is because it skips the redeem step when gasPrice==0 and/or maxGas==0.  This can probably be fixed by doing an auto-redeem in every case when fees are disabled.

This doesn't affect Rinkeby-net or Arbitrum One because they have fees enabled.

edfelten:
This makes some changes to simplify the semantics of gas, and eliminate some non-intuitive behaviors.

First, when a sub-tx finishes, it verifies that the sub-tx's unused gas would be enough to pay for the net storage allocation done by the sub-tx.  If not, it reverts the sub-tx, as if it had run out of gas.  This ensures the invariant that a sub-tx can't deplete its parent's gas by more than the gas amount given to the sub-tx.

Second, the EVM GAS instruction now fully accounts for the storage cost that the current tx would incur if it finished immediately.  If the current tx has been a net allocator of storage, the number of cells allocated, multiplied by the current ArbGas cost per storage cell, is subtracted from the remaining gas, to determine the value returned by the GAS instruction.  If the result would be negative, the tx is immediately treated as out-of-gas.  This causes the GAS instruction to have more intuitive behavior.


rachel-bousfield:
Our gas-charging model has evolved over time via other PRs, so this one is out of date with our current thinking

PR #553, opened 2021-07-12
Rigorous Assignability Rules

RachelFranks:
Updates the language specification to include complete rules for
the assignability and equality of types.

After converging on the ideal set of rules, we'll update the
typechecker to carry them out.

### DIFF
--- a/doc/MiniLanguageTutorial.md
+++ b/doc/MiniLanguageTutorial.md
@@ -153,7 +153,7 @@ The compiler uses often uses type inference to infer the types of variables from
 
 Two types are equal if each is assignable to the other.
 
-A type definition, whether imported or defined locally, does not create a new type but rather creates a new name for the same underlying type.
+A type definition `type A = B`, whether imported or defined locally, creates a new type `A` to which `B` is assignable.
 
 ## Castability
 
--- a/doc/MiniLanguageTutorial.md
+++ b/doc/MiniLanguageTutorial.md
@@ -1,6 +1,6 @@
 # Mini language tutorial
 
-[version of June 5th, 2020]
+[version of July 12th, 2021]
 
 Mini is a programming language and compiler designed for writing code for the Arbitrum Virtual Machine (AVM) platform.  The language is simple but has some features customized for AVM.  This tutorial will tell you what you need to know to write Mini programs.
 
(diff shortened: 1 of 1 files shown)


doc/MiniLanguageTutorial.md
```diff
@@ -132,51 +132,28 @@ Mini has the following types:
...
+A value of type `V` is assignable to storage of type `S` if and only if any of the following are true:
 
-Every atomic type is equal to itself.
+* `S` and `V` are the same atomic type
+* `S` and `V` are options, and the inner type of `V` is assignable to the inner type of `S`
+* `S` and `V` are tuples with the same number of fields, and each field of `V` is assignable to the corresponding field of `S`
+* `S` and `V` are arrays, and the member type of `V` is assignable to the member type of `S`
+* `S` and `V` are fixed-size arrays with the same size, and the member type of `V` is assignable to the member type of `S`
+* `S` and `V` are structs with the same number of fields, and each field of `V` has the same name as and is assignable to that of `S`
+* `S` and `V` are maps, and the key and value types of `V` are correspondingly assignable to that of `S`
```
edfelten, line 145: Shouldn't maps be treated like a function from keyType to valueType, therefore map<K1, V1> is assignable to map<K2, V2> if K2 is assignable to K1 and V1 is assignable to V2?

doc/MiniLanguageTutorial.md
```diff
@@ -1,6 +1,6 @@
 # Mini language tutorial
 
-[version of June 5th, 2020]
+[version of July 12th, 2021]
```
edfelten, line 3: Maybe better to take this out?  Since the file is tracked in git, the version information should be available from git--which will be more reliable than a manually maintained version date.

edfelten, line 3: Maybe better to take this out?  Since the file is tracked in git, the version information should be available from git--which will be more reliable than a manually maintained version date.

edfelten (commented):
A couple of minor suggestions.

doc/MiniLanguageTutorial.md
RachelFranks: Removed -- it's not particularly clarifying and it's easy to forget about it

edfelten (approved):
Looks good.

PR #557, opened 2021-07-16
Simplify Console Colors

rachel-bousfield:
- Adds color functions for easy string coloring
- Replaces CompileError colors with proper console color struct
- Removes all instances of unwieldy color use via the new functions


rachel-bousfield:
Solving the merge conflicts resulted in no change

PR #559, opened 2021-07-19
Fix extcode* instruction semantics in constructor

### LINKED ISSUE #558: EXTCODESIZE evm opcode in contract constructor doesn't return 0
Calling the `EXTCODESIZE` evm opcode inside of a contract constructor on Arbitrum appears to behave differently than on other EVM chains - returning the size of the code being deployed instead of `0`.

According to the EVM spec in the [Ethereum yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf), it should return `0` during contract initialization:

> During initialization code execution, `EXTCODESIZE` on the address should return zero, which is the length of the code of the account while `CODESIZE` should return the length of the initialization code (as defined in H.2).

I've created a repo with a contract that can be deployed on other EVM chains but not on Arbitrum: https://github.com/ren-forks/arbitrum-extcodesize-issue

edfelten:
The EVM EXTCODESIZE, EXTCODEHASH, and EXTCODECOPY instructions have special behavior if called on an address for which a constructor is currently running. In that scenario, they're supposed to behave as if the account has no code, even though calling CODESIZE, CODEHASH, or CODECOPY would show non-empty code.  This fixes the behavior of the EXT family instructions to match the EVM spec.

Fixes #558


edfelten:
Closing this because we're going to merge these changes earlier.

PR #560, opened 2021-07-19
Fix EXTCODE* EVM instructions in constructors

edfelten:
According to the EVM spec, the EXTCODE* instructions, when invoked on the address of a contract whose constructor is currently running, should behave as if the address has no code.  Previously these used the code of the constructor, which is the specified behavior for the CODE* instructions.

Fixes #559


src/evm/mod.rs
```diff
@@ -1054,6 +1054,33 @@ pub fn evm_direct_deploy_add(log_to: Option<&Path>, debug: bool) {
     }
 }
 
+pub fn _evm_test_extcodesize_of_constructor(log_to: Option<&Path>) {
```
RGates94, line 1057: I would recommend placing this within a test file so the compiler won't ignore it if it becomes unused.

RGates94 (commented):
Mostly looks good, one small suggestion.

arb_os/evmOps.mini
```diff
@@ -309,34 +312,19 @@ public impure func evmOp_extcodecopy(addrAsUint: uint, memAddr: uint, codeOffset
...
-                code,
-                codeOffset,
-                memory,
-                memAddr,
-                nbytes
-            );
-            if (evmCallStack_setTopFrameMemory(updatedMemory,)) {
-                return;
-            }
+        let (codeArr, codeSize) = xif (evmCallFrame_constructorIsOnStackForAddress(topFrame, addr)) {
```
PlasmaPower, line 315: I believe `codeSize` is actually an offset, not a size, and that we should have the result of the if statement just be `codeArr` and use the `codeOffset` parameter as-is. I'm assuming reading from a non-zero offset into a zero array is just zeroes.

arb_os/evmOps.mini
```diff
@@ -350,11 +338,15 @@ public impure func evmOp_extcodehash(addrAsUint: uint) -> bytes32 {
...
-            return keccak256(code, 0, bytearray_size(code));
+        return xif (evmCallFrame_constructorIsOnStackForAddress(topFrame, addr)) {
+            bytes32(0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470)  // hash of empty code
         } else {
-            return bytes32(0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470);
-        }
+            xif let Some(code) = account_getCode(acct) {
+                keccak256(code, 0, bytearray_size(code))
+            } else {
+                bytes32(0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470)  // hash of empty code
```
PlasmaPower, line 347: I believe mini has constants, now that this is used twice, should this be moved to a constant?

PlasmaPower (changes requested):

arb_os/evmOps.mini
```diff
@@ -350,11 +338,15 @@ public impure func evmOp_extcodehash(addrAsUint: uint) -> bytes32 {
         if (account_isEmpty(acct)) {
             return bytes32(0);
         }
-        if let Some(code) = account_getCode(acct) {
-            return keccak256(code, 0, bytearray_size(code));
+        return xif (evmCallFrame_constructorIsOnStackForAddress(topFrame, addr)) {
+            bytes32(0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470)  // hash of empty code
```
PlasmaPower, line 342: Should this return the empty code hash or `0`? EIP-1052 isn't clear on this. It doesn't seem like `account_isEmpty` will ever trigger in this case, assuming `contractInfo` is set at this point.

src/evm/mod.rs
edfelten: Changed this to #[cfg(test)] and removed initial underscore from its name.

arb_os/evmOps.mini
edfelten: Good idea; done.

arb_os/evmOps.mini
edfelten: Done.

arb_os/evmOps.mini
edfelten: Agree that the EIP is unclear.  It seems safer to leave this as is.

PlasmaPower (dismissed):
I've confirmed EXTCODEHASH on a constructor works the same way in Eth

arb_os/evmCallStack.mini
```diff
@@ -1102,6 +1108,7 @@ public impure func evmCallStack_callRetryableTxAndReturn(rtx: RetryableTx) -> (b
             bytearray_size(rtx.calldata),
             0,
             0,
+            false,
```
hkalodner, line 1111: Confirming that the behavior of retryables sent to the 0 address is not to call it as a constructor. We should clarify the documentation if it isn't already clear

arb_os/evmCallStack.mini
```diff
@@ -1306,6 +1313,19 @@ public impure func evmCallStack_oldestCallFrame() -> option<EvmCallFrame> {
     };
 }
 
+public func evmCallFrame_constructorIsOnStackForAddress(frame: EvmCallFrame, addr: address) -> bool {
```
hkalodner, line 1316: I'm wondering if there are any edge case deviations here involving nested CREATE2s

CREATE2 contract A
    CREATE2 contract A
    EXTCODESIZE contract A
    SELFDESTRUCT contract A

I think in both ethereum and arbitrum the operations would be valid, but with this implementation in arbitrum  the EXTCODESIZE would return 0 rather than the size of the code

edfelten: I haven't tested it, but I think that if there is a CREATE2 for address A, with another CREATE2 for the same address A nested inside it, the second/inner CREATE2 will fail because it will see that there is already a contract at A.

PlasmaPower: That should be the case from what I remember of the code, as CREATE2 instantly populates contractInfo, but will fail if contractInfo already exists.

arb_os/evmCallStack.mini
edfelten: Yes, a retryable sent to address zero will be treated as a tx to address zero, not as a deploy.  There shouldn't be any way to create a retryable that is a deploy.

PlasmaPower (approved):
LGTM

PlasmaPower (dismissed):
LGTM

arb_os/evmCallStack.mini
hkalodner: We should clarify that there's no way to use a retryable to deploy a contract since I think that's likely non-intuitive

hkalodner (approved):
LGTM

edfelten:
Superseded by #561

PR #562, opened 2021-07-24
Add regression test for gasleft with delegatecall issue

edfelten:
This adds a regression test for an issue that was causing erroneous results for `gasleft` with nested delegatecalls.  The issue is no longer current, so the only change is addition of a regression test to prevent it from coming back.


RachelFranks (approved):
Looks good to me

PR #563, opened 2021-07-25
Clear the AVM stack before invoking constructor

edfelten:
Clear the AVM stack before invoking a constructor. Previously the constructor was called via a Mini function call, which left a return address on the stack.


PlasmaPower (approved):
LGTM

PR #564, opened 2021-07-26
Charge for storage when adding to function decompression table

edfelten:
When a tx expands the size of an aggregators function decompression table, this adds a charge of 3 storage for each added table item.


PlasmaPower (approved):
LGTM

PR #565, opened 2021-07-26
Preserve account balance across create/create2

edfelten:
This preserves the state of an account, except for contract-related parts, across create or create2 operations that create contracts at the account's address.


PlasmaPower:
What if the constructor has callvalue and/or moves balance to/from its address? I think we need to re-fetch the account afterwards.

PlasmaPower (approved):
LGTM

PR #567, opened 2021-07-26
Charge for memory size

edfelten:
This adds functionality to charge EVM programs for their use of EVM memory. We use the same basic schema as Ethereum, where a tx is charged proportionally to the square of the largest address read or written in its memory.  Specifically, if the largest such index is X, then we charge `X*X / Denom` ArbGas, where Denom is a chain parameter that is initially 100_000. This is about 5% of what Ethereum would charge for the same memory activity.


PlasmaPower (changes requested):
We need safemath in a bunch of places in this PR, such as gas computation, and maybe even adding the offsets and lengths. Perhaps saturating math would be simpler to use here.

One other side note, we need to be careful with the `length > 0` checks that we actually ignore 0 length copies and such at high offsets.

PlasmaPower:
You were definitely right about this being the simpler approach though, covering evmOps is a much more straightforward surface area.

PlasmaPower (approved):
LGTM

PR #569, opened 2021-07-27
Upload buffer support

RachelFranks:
Adds unmarshalling code in ArbOS and buffer upload code in rust that allows us to support non-empty buffers. This also includes a minimal patch from future PR #568 to enable string support and multi-line comments.


src/compile/codegen.rs
```diff
@@ -817,12 +817,32 @@ fn mavm_codegen_expr<'a>(
...
         TypeCheckedExprKind::NewBuffer => {
             let opcode = Opcode::AVMOpcode(AVMOpcode::NewBuffer);
             code.push(Instruction::new(opcode, None, debug));
             Ok((label_gen, code, num_locals))
         }
+        TypeCheckedExprKind::Quote(bytes) => {
+            code.push(create!(
+                Noop,
+                Value::new_tuple(vec![
+                    Value::Int(Uint256::from_usize(2 * bytes.len())),
```
edfelten, line 840: Why do we multiply bytes.len by 2 here?

edfelten (commented):
One question. Otherwise, looks good.

src/compile/codegen.rs
RachelFranks: Oops, that's wrong. I've committed a fix

edfelten: Thanks.  Looks good.

PR #571, opened 2021-07-27
No refund for storage deallocation

edfelten:
This changes the charging schema for EVM storage, so that a contract gets charged for allocating new storage (setting a cell from zero to non-zero) but doesn't get refunded for deallocating (setting from non-zero to zero).

This avoids various corner cases in the charging / gas left implementations.


rachel-bousfield:
We've changed the refund mechanism & model a lot since this PR was made

PR #573, opened 2021-07-29
View and Write

RachelFranks:
Removed `impure` keyword and concept in favor of:
- `view` for when a series of instructions depends on global state
- `write` for when a series of instructions writes to global state

This also adds
- Warnings for when `view` and `write` are applied too liberally
- Errors for when `view` and `write` should be present but aren't
- Changes to ArbOS to ensure functions are properly marked `view` and/or `write`
- Test case to ensure appropriate behavior around function pointers


src/gen_code/mod.rs
```diff
@@ -1,4 +1,9 @@
+//
+// Copyright 2020, Offchain Labs, Inc. All rights reserved.
```
edfelten, line 2: This should probably list the year as "2020-2021" because it appears to contain code written in both years.

src/mavm.rs
```diff
@@ -977,27 +980,42 @@ pub enum AVMOpcode {
 }
 
 impl Opcode {
-    pub fn is_pure(&self) -> bool {
+    pub fn is_view(&self, _: &TypeTree) -> bool {
+        match self {
+            Opcode::AVMOpcode(AVMOpcode::Inbox)
+            | Opcode::AVMOpcode(AVMOpcode::InboxPeek)
+            | Opcode::AVMOpcode(AVMOpcode::Send)        // 2 sends can't be reordered even though no state is read
```
edfelten, line 987: Shouldn't the send opcode be write but not view?  (The comment about reordering doesn't have implications for view, because it's safe to reorder two view operations.  But write operations can't be reordered.

src/mavm.rs
```diff
@@ -977,27 +980,42 @@ pub enum AVMOpcode {
...
+            | Opcode::AVMOpcode(AVMOpcode::Rpush)
+            | Opcode::AVMOpcode(AVMOpcode::PushInsn)
+            | Opcode::AVMOpcode(AVMOpcode::PushInsnImm)
+            | Opcode::AVMOpcode(AVMOpcode::ErrCodePoint)
+            | Opcode::AVMOpcode(AVMOpcode::ErrPush)
+            | Opcode::AVMOpcode(AVMOpcode::PushGas)
+            | Opcode::AVMOpcode(AVMOpcode::Sideload)    // this is special-cased, so we can't reorder these
```
edfelten, line 994: I don't think the reordering comment is relevant here.  (It is relevant below, where Sideload is listed as write.)

edfelten (changes requested):
Just a few minor comments.

src/gen_code/mod.rs
RachelFranks: Good idea, I'm updating the copyright notice

src/mavm.rs
RachelFranks: You're right, it's just `write`. I'm changing this

src/mavm.rs
RachelFranks: Yeah, it's not. I've removed it

PR #575, opened 2021-08-02
Closures

RachelFranks:
- Added closures + test cases to mini (see notes)
- Fixed replica functions check regression and added regression test
- Removed the Noop() AST node type
- Added pretty-printing tools that make debugging codegen intensive things (like closures) much easier

Notes
- Closures are fully typechecked, including unused assignment checking, flow analysis, etc. This includes across the capture boundary.
- The number of captures has no bearing on type -- they are broadly interchangeable via a dynamic unpacking algo at runtime. This means that a function is indistinguishable from a closure with no captures and can be used in its place. Hence, there's no type difference between a closure and a func.
- Capture costs are amortized -- using a closure over and over as in the case of iterating over a map is more efficient than iterating with a function pointer like we currently do in ArbOS.
- You're safe against accidental capturing of a global variable because this is only allowed when you declare the closure as `view` and/or `write`. (The view-write system is exactly the same as in funcs).
- Global state is *not* captured -- to get the same effect as a regular capture you must pass in the value of a global via a local capture. This we can change if need be, but the idea for why we do this is that one might want to edit global state during a `write` closure.
- You can use `closure` in place of `func` when defining types. It's just syntactic sugar since no distinction is needed between them.


src/mavm.rs
```diff
@@ -37,7 +40,7 @@ impl Label {
 
     pub fn avm_hash(&self) -> Value {
         match self {
-            Label::Func(sid) => Value::avm_hash2(
+            Label::Func(sid) | Label::Closure(sid) => Value::avm_hash2(
```
edfelten, line 43: Should the hash of a closure include the captured values?  Or maybe include the pre-built call frame (which includes the captured values)?

(And if we change this, do we want to change how functions are hashed, so that a function is equivalent to a closure with no captures?)

src/mini.lalrpop
```diff
@@ -88,18 +97,17 @@ BodyDecl: TopLevelDecl = {
 
 TypeDecl: TypeDecl = {
 	"type" <Ident> "=" <Type> => new_type_decl(<>),
+    "type" <Ident> "=" <Type> ";" => new_type_decl(<>),
```
edfelten, line 100: The lack of semicolons after type definitions is a historical artifact. It seems more consistent going forward to require a semicolon.  Should we do that?

edfelten (changes requested):
Looks good overall.  I have a few detailed questions to discuss.

src/mini.lalrpop
RachelFranks: I think we should just go ahead and require a semicolon. I'll commit a change that removes the top rule and fixes ArbOS

src/mavm.rs
RachelFranks: So this is something that's a bit interesting. As it currently stands, func & closure label hashes are not unique. In fact, the vast majority of funcs share hashes with each other. This is because we are currently assigning them based on the local `StringId`of the func. After `xlating` they are differentiated, but this scheme is dangerous and hacky.

In the next PR #577, I make func & closure label mavm hashes based on the rust-hash of their file & names, ensuring they are unique across all stages of compilation and between versions of ArbOS. I think we should adopt the new scheme, accepting this one as a temporarily equivalent to the first that also happens to include closure labels.

PR #576, opened 2021-08-02
Add gas estimation tests

edfelten:
Add tests of gas estimation functionality.  The tests do gas estimation through preferred aggregator and non-preferred aggregator, then runs the transaction again for real, with gas amount based on the estimate.

### DIFF
--- a/src/evm/abi.rs
+++ b/src/evm/abi.rs
@@ -197,6 +197,52 @@ impl AbiForContract {
         ))
     }
 
+    #[cfg(test)]
+    pub fn estimate_gas_for_function_call(
+        &self,
+        func_name: &str,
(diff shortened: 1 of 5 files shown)


PlasmaPower:
This should test that `units * prices = paid`, which would catch the issue fixed in https://github.com/OffchainLabs/arb-os/pull/578

edfelten:
> This should test that `units * prices = paid`, which would catch the issue fixed in #578

It does that, starting at line 618 of minitests/mod.rs

for i in 0..4 {
        assert_eq!(
            estimate_fee_stats[0][i].mul(&estimate_fee_stats[1][i]),
            estimate_fee_stats[2][i]
        );
    }

PR #577, opened 2021-08-03
Remove old import export system

RachelFranks:
- Completely removes the old import export system
- Funcs and closures now have deterministic labels, making them module agnostic
- Fixes a bug regarding `FileInfoChart` aggregation
- Ensures all builtins have a `StringId` to reduce special casing


edfelten (approved):
Looks good.

PR #578, opened 2021-08-04
Record network fee in fee stats prices in new chain params system

PlasmaPower:
This fixes arbitrum repo fee tests on this branch. CI will fail as on the base branch because the upgrade test uses the wrong chain params.


edfelten (approved):
Good catch!

PR #580, opened 2021-08-08
Return different result code on out-of-gas

### LINKED ISSUE #579: Return a different result code when a top-level tx runs out of gas
Currently ArbOS returns the ordinary revert result code (1) when a transaction runs out of gas.  Returning a different result code would be useful to other tooling, allowing tooling to tell the difference between a tx that runs out gas vs one that fails for another reason.

edfelten:
Create a new transaction result code (16) to indicate that the transaction ran out of gas while executing. Previously ArbOS returned the result code for an ordinary revert (1) in this case.

Fixes #579


arb_os/errorHandler.mini
```diff
@@ -66,6 +70,14 @@ public view write func errorHandler() {
     } else {
         // Application code must have hit an error.
         inErrorHandler = false;
+        let _ = evmCallStack_returnFromCall(
```
RachelFranks, line 73: Now that `evmCallStack_returnFromCall` may resolve the error handler, the `debug((664, errInfo = 0));` from `evmCallStack_callHitError` may not occur. This shouldn't matter since we aren't consuming this on the node's side

RachelFranks (approved):
Looks good to me, merging

PR #581, opened 2021-08-09
Better compiler defaults

RachelFranks:
By default, `-o` now pipes to /dev/null
```
mini compile arb_os
mini compile arb_os -o /dev/null
```

By default, `-c` looks for a `constants.json` in the same directory
```
mini compile arb_os
mini compile arb_os -c arb_os/constants.json
```


edfelten (approved):
Looks good.

PR #585, opened 2021-08-10
Upgrade 4

RachelFranks:
We should probably merge in each upgrade as they are deployed. Either way, I've opened this PR so that the node tests will run.


RachelFranks:
Upgrade 4 is ready for deployment [upon approval from the node team](https://github.com/OffchainLabs/arbitrum/pull/1489)

hkalodner (approved):
High level approval that the node integration works. All particular components have been fully reviewed

PR #588, opened 2021-08-10
Move codegen to library

edfelten:
This moves the EVM-to-AVM translator from the main ArbOS directory into the standard library.  This makes sense now that the translator doesn't use global variables (except to cache immutable values).


RachelFranks (approved):
Looks great to me -- this will probably be the last PR for upgrade 7

PR #589, opened 2021-08-11
Default initialize globals

RGates94:
This causes globals to be initialized with their type's default value.


RachelFranks:
We now have a version of this in #607 that takes into account nominal types and fixes a few correctness problems in our current value defaults. The work needed to make this work pre `#function-centric-codegen` isn't worth it, since global handling in that branch on is easier to test and reason about

PR #590, opened 2021-08-12
No more beeps

RachelFranks:
- Makes pretty printing buffers as utf-8 exclude control characters
- Pretty printing is no longer the default for buffers for correctness reasons
-- you must call `pretty_print` on an `mavm:Value` to make nested buffers human-readable
- Adds string-based error codes to a few test files
- Adds prettier import error


edfelten (approved):
Looks good.

PR #594, opened 2021-08-14
Function centric codegen

RachelFranks:
Makes codegen function-centric so that codegen is module-agnostic.
This makes future optimizations possible and enables other refactors to codgen.

This PR also
- Reduces the number of backward jumps from 279 to 11 for a 0.5% performance gain
- Reduces the number of AVM instructions in ArbOS by 12.88%
- Multi-threads codegen
- Creates .dot graph files for callgraph viewing
- Globals can now have attributes & debug info
- Refactors unused function procedure to not generate code for stdlib funcs
- Refactors `label_gen` to be less stateful
- Removes lifetime parameters from codegen
- Removes all code relocation & simplifies linking
- Adds the `string` datatype
- Reduces special-casing around closures
- Improves import error messages and logic


RachelFranks:
The nightly tests are failing due to a [bug in the rust compiler](https://github.com/rust-lang/rust/issues/88043).

RGates94 (approved):
LGTM

PR #595, opened 2021-08-17
Upgrade 7

RachelFranks:
#588 is probably the last PR we should include in upgrade 7. I'll merge it in after I review it.

Will rebase when `upgrade 6` is added.


RachelFranks:
#614 became our actual `upgrade-7`, so I'm closing this PR.

PR #607, opened 2021-09-03
Default global values

RachelFranks:
- Default initializes global values
- Adds debugging tools for globals, their layouts, and access patterns
- Fixes default value bugs (please review these carefully since they were originally wrong)
- Adds a tuple-surgery method that lets you conditionally replace nested values deep within tuples


src/compile/ast.rs
```diff
@@ -913,66 +913,75 @@ impl Type {
...
-                let mut vals = Vec::new();
-                let mut is_safe = true;
-                for field in fields {
-                    let (val, safe) = field.tipe.default_value();
-                    vals.push(val);
-                    is_safe = is_safe && safe;
-                }
-                (value_from_field_list(vals), is_safe)
+            Type::Nominal(..) => {
+                let tipe = self.rep(type_tree).unwrap_or(Type::Any);
```
RGates94, line 972: This should throw an error if the rep function fails. As it stands it would not be type safe on read, and that could lead to a silent error that's only discovered at run time.

src/compile/ast.rs
```diff
@@ -913,66 +913,75 @@ impl Type {
...
+                tipe.default_value(type_tree)
             }
-            Type::Map(_, _) | Type::Func(_, _, _) | Type::Nominal(_, _, _) => {
-                (Value::none(), false)
+            Type::Generic(_) | Type::GenericSlot(_) => {
+                // This can happen with builtins, but we plan on refactoring all builtins out.
+                // In the future we'll make this panic since you should always know the specified value
+                // for whatever generic type you're getting the default for.
+                Value::none()
```
RGates94, line 979: Unless there is some rationale I am missing, it might be better to keep a safety marker in the return type if unsafe defaults are still possible. We can warn if something like this is present on a type if this marker is there.

RGates94 (changes requested):
Looks good overall, a couple small things.

src/compile/ast.rs
RachelFranks: Nice catch, I was debugging something and forgot to change that back to a panic. I chose to panic since there really shouldn't be a way to make it far enough to codegen & linking with an unresolvable type -- and if we find a way to do that, we'd want to do this check while typechecking the AST.

src/compile/ast.rs
RachelFranks: The rationale is that, although this function used to return a safety marker, it was always underscored and discarded. `NewArray`'s codegen was the only caller and, since the type tree is lost at this point, we've been filling it with empty tuples. Now we do have `link()` as the additional caller, but, since globals always have a specialized version of their generic types, you can't have any `Generic` or `GenericSlot` sub-types. So the only scenario where this would cause a bug is one where we already have prior to this PR.

You're hitting on something good here, though, which is that we should move to not using builtins since their presence edge-cases the safety guarantees of constructs like generics. Once we remove `newarray` in specific, we can make this function panic on these two types.

RGates94 (approved):
LGTM

PR #608, opened 2021-09-04
Basic blocks

rachel-bousfield:
- [x] Add basic block representation
- - [x] Refactor meta-opcodes to never use immediates
- - [x] Delay func calls until linking
- [x] Supporting prior-pass changes
- - [x] Remove the StructInitializer & DotRef AST nodes
- - [x] Remove `xif` and `xloop` + beautify grammar
- - [x] Remove most builtins from codegen
- - [x] Remove `StatementKind::Asm`
- - [x] Make `None` not require `None<Type>` for non-compares
- - [x] Flowchecker return expressions
- - [x] Move codegen typechecking to typecheck
- - - [x] Warn/Error on dropped expressions
- - - [x] Check that `asm()` codeblocks return a value
- - - [x] Ensure loop codeblocks don't return a value
- - - [x] Typecheck `break` statements (moved to future PR)
- [x] Rewrite codegen to be SSA
- - [x] Refactor codegen to be single-function
- - [x] Remove locals counting for frame sizes
- - [x] Fix outstanding codegen bugs
- - - [x] 9-way mixed-let
- - - [x] Folding tuple ops
- - - [x] func call & asm stack smashing
- - [x] New intermediate representation
- - - [x] Single static assignments
- - - [x] Phi nodes
- - [x] Conditional compilation
- [x] Register coloring
- - [x] Graph-contract phi-joins
- - [x] Make slot assignments
- - [x] Update closures


src/compile/codegen.rs
```diff
@@ -4,104 +4,156 @@
...
+    fn set_local(&mut self, local: StringId, slot: SlotNum) {
+        let len = self.scopes.len();
+        self.scopes[len - 1].locals.insert(local, slot);
+    }
+
+    fn shadow(&mut self, local: StringId, slot: SlotNum) {
+        let len = self.scopes.len();
+
+        // for the first time shadowing, we save the old value if it exists
+        if let Some(old) = self.scopes[len - 1].locals.get(&local) {
```
RGates94: We should consider using .get() or checking the len > 0 to avoid the possibility of a panic.

rachel-bousfield: This is outdated, I recommend pulling the latest version. We should still panic because not having a scope means something is very wrong.

src/mini.lalrpop
```diff
@@ -90,24 +87,85 @@ PathDecl: Vec<String> = {
...
         let write = qualifiers.contains(&"write");
         let debug = DebugInfo::new(file_info.location(BytePos::from(lno), filename), attribs.unwrap_or_default());
-        let name = stringtable.name_from_id(i).clone();
+        let name = string_table.name_from_id(i).clone();
+
+        // Generalize over the generic parameters
+        let params: Vec<_> = g.into_iter().flatten().collect();
+        let generic_ids = check_generic_parameters(params.clone(), &string_table)?;
+        args.iter_mut().for_each(|arg| arg.tipe = arg.tipe.make_generic(&generic_ids));
+        if let Some(ref mut tipe) = ret {
```
RGates94: This is simpler as `Some(tipe) = &mut ret`

rachel-bousfield: Sure, we can use that style. I've pushed a change

edfelten (approved):
Thanks for the PR, and for the reviewing guide.  This looks good me.

RGates94 (approved):
This looks good to me.

PR #610, opened 2021-09-08
Blake2f instruction errors if steps > 0xffff

edfelten:
Make the `blake2f` AVM instruction generate an error if the step count is greater than `0xffff`.  This is a safer way to deal with unexpected inputs, and consistent with the implementation being done in the C++ AVM.

### DIFF
--- a/src/run/blake2b.rs
+++ b/src/run/blake2b.rs
@@ -47,30 +47,24 @@ impl Clone for Blake2b {
     }
 }
 
-pub fn blake2bf_instruction(buf: Buffer) -> Buffer {
+pub fn blake2bf_instruction(buf: Buffer) -> Option<Buffer> {
     let mut b = [0u8; 213];
     for i in 0..b.len() {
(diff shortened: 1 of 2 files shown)


src/run/blake2b.rs
```diff
@@ -47,30 +47,24 @@ impl Clone for Blake2b {
     }
 }
 
-pub fn blake2bf_instruction(buf: Buffer) -> Buffer {
+pub fn blake2bf_instruction(buf: Buffer) -> Option<Buffer> {
     let mut b = [0u8; 213];
     for i in 0..b.len() {
         b[i] = buf.read_byte(i as u128);
     }
     b[212] = b[212] & 0x1u8;
```
tsahee, line 55: the current implementation for node (and onestepproof) also treat as error:
* buffer with more than 213 bytes
* if byte 212 ("f" flag) is > 1

tsahee (approved):

PR #611, opened 2021-09-10
Retryables builtin interface refactor

fredlacs:


hkalodner (dismissed):
LGTM, but would be good to have somebody from the arb-os team confirm that this doesn't create issues for the build system there

rachel-bousfield (dismissed):
The build system still works, as mini's dependencies on `contracts/artifacts/arbos/builtin/*.sol/*.json` are still `make`-able via `yarn build`

rachel-bousfield (approved):
The build system looks great

PR #612, opened 2021-09-10
Colored-any Generics

RachelFranks:
- Adds generics & tests
- Fixes bugs during type check
- Adds a work-in-progress generic standard library


RGates94 (commented):
While this PR does a good job at implementing some forms of generics, it seems like recursive generic types aren’t working right now. For example, this implementation of a generic queue:
```rs
type queue<T> = (T, option<queue<T> >);

write func main() {
    return;
}

func dummy() -> queue<T> {
    return unsafecast<queue<T> >(0);
}
```

This produces the following compiler output:
```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
```

RachelFranks:
You forgot to declare `T` as a generic arg.

```rs
type queue<T> = (T, option<queue<T> >);

write func main() {
    return;
}

func dummy<T>() -> queue<T> {
    return unsafecast<queue<T> >(0);
}
```

The reason for the infinite loop was due to a bug in the error printing function, which tried to print the error out recursively forever. I've fixed this, as well as addressed an edge case with unsafe cast. There's a new `queue.mini` test case to prevent a regression.

Please let me know if you hit any other issues with this. I confident the approach is general but there could be some rough spots, which I'm happy to fix.

src/compile/ast.rs
```diff
@@ -456,42 +618,47 @@ impl Type {
...
-                    self.get_representation(type_tree),
-                    rhs.get_representation(type_tree),
-                ) {
+            Type::Nominal(_, _, _) => {
+                if let (Ok(left), Ok(right)) = (self.rep(type_tree), rhs.rep(type_tree)) {
                     if seen.insert((left.clone(), right.clone())) {
-                        left.assignable(&right, type_tree, seen)
+                        let result = left.assignable(&right, type_tree, seen.clone());
+                        if !result {
+                            dbg!(&left, &right);
```
RGates94, line 658: This debug print should probably be removed, as it is hard to tell what it means without looking at the source code.

RGates94 (changes requested):
Looks mostly good just one small thing I noticed.

src/compile/ast.rs
RachelFranks: Ah nice catch, accidentally left that in while debugging

RGates94 (approved):
Looks good.

PR #615, opened 2021-09-14
Proportionally allocate storage gas to callees

edfelten:
This changes how gas accounting for computation vs storage is managed, for newly deployed contracts.

Here's a summary of how storage charging works, after this update. 
* As before, the gas for a top-level tx is divided into "all-purpose gas" which can cover computation or storage, and "storage-only gas" which can only pay for storage. The tx gas limit is a limit on all-purpose gas.  If a tx includes less than the tx gas limit (after deducting L1 fees), then all of its gas goes to all-purpose and none to storage-only.  If a tx includes more than tx gas limit (after L1 fees) then it gets all-purpose gas equal to the tx gas limit and the rest of its gas goes to storage-only.
* If a call allocates storage, the gas cost of that storage allocation is first deducted from storage-only gas, and if the storage-only gas is insufficient then the remainder is deducted from all-purpose gas.
* Computation uses all-purpose gas. A call reverts as out-of-gas when its all-purpose gas is all gone.
* Solidity `gasleft()` and the equivalent EVM `GAS` instruction return the current amount of all-purpose gas.  A new ArbSys method, `getStorageGasAvailable`, returns the current amount of storage-only gas.
* Calls that specify a gas amount to give to the callee are specifying an amount of all-purpose gas.
* For calls from contracts deployed after this PR is live on-chain, a call that gives a fraction `f` of the caller's (all-purpose) gas will automatically also give the same fraction `f` of the caller's storage-only gas.   (For calls from older contracts, all of the caller's storage-only gas is given to the callee, for backward compatibility.)
* If a transaction reverts, it is charged for the all-purpose gas it used, but all of the storage-only gas it was given is returned to the caller (because the storage allocations were reverted).
* There is no refund for deallocating storage.


RachelFranks:
Initial testing shows the following node tests fail in both the clean start and upgrade scenarios
- TestNonAggregatorFee
- TestRetryableFee
- TestArbOSFees

This is likely expected, as we're changing fees here. Importantly, all non-fee tests pass

arb_os/evmCallStack.mini
```diff
@@ -869,7 +889,8 @@ public view write func evmCallStack_returnFromCall(
         evmTracer_pushReturnRevert(
             resultCode,
             returnData,
-            resumeInfo.givenGas - leftoverGas,
+            (resumeInfo.givenAPGas + resumeInfo.savedAPGas - leftoverAPGas)
```
RachelFranks, line 892: I don't think this should have the saved gas here

RachelFranks (changes requested):

PR #616, opened 2021-09-14
Allow anyone to call the view methods in ArbOwner

edfelten:
This allows anyone to call the view (non-mutating) methods of the ArbOwner precompile.  Non-view methods still require that the caller has owner privileges.

These methods are useful for things like chain monitoring tools.  This is not a confidentiality breach because the information was already assumed public and has been available to anyone running a node.


rachel-bousfield:
Moved to Nitro

PR #618, opened 2021-09-17
Rinkeby Upgrade 8 Part 2

RachelFranks:
Upgrades Rinkeby to include new safety measures

This upgrade includes the following PRs
- #619

PR #620, opened 2021-09-20
Mainnet upgrade 8

RachelFranks:
This mainnet upgrade combines the two on Rinkeby with special bridging code
- #618 
- #617


edfelten (approved):
Looks good.

PR #621, opened 2021-09-20
Separate approval of upgrade from triggering it

edfelten:
This separates approving an upgrade, which can only be done by the chain owner, from actually triggering an already-approved upgrade, which anyone can now do.

Because triggering an upgrade causes the current transaction to end in an unusual way, we require it to be done in a "top-level" tx that was submitted directly from L1.  By allowing upgrades to be approved by the chain owner without immediately triggering them, we allow the chain owner to be an L2 contract.  This opens up more options for using, e.g., multi-sig and time-delay functionalities to control the upgrade capability.  And it help enable tooling to generate and authorize upgrades that is simpler and more error-resistant.

The approver provides a timestamp after which the upgrade will no longer be triggerable.

Triggering an already approved (and not timed out) upgrade can now be done by anyone, so it is easy to do directly from L1, which is still required.


edfelten:
This fails CI because it declares a constant that isn't used.  That's for good reason. It's needed to support the test that upgrades from the previous version.

rachel-bousfield:
Moved to Nitro

PR #623, opened 2021-09-21
Version Number and Branch Organization

RachelFranks:
Under the new scheme
- `master` becomes the exact state of ArbOS as deployed on mainnet.
- `develop` becomes the next upgrade we intend to deploy
- Each deployment will have a corresponding `upgrade-v` branch where `v` is the ArbOS version number

Version numbers change as follows
- Upgrades will be denoted by the _unique_ ArbOS version number in the `constants.json`
- GitHub releases, which will start at `0.9`, will be released with public announcements


edfelten (approved):
Approving this as an administrative change, to merge the active deployed version into master.

PR #624, opened 2021-09-22
Refund gas charged for storage allocation, if deallocate within the same tx

### LINKED ISSUE #622: Don't charge for storage allocation when deallocated in same tx
Upgrade 9 fixed some issues with storage accounting, but it had one effect that we want to reverse. In particular, we want to make it so that if a transaction allocates a storage cell (i.e. sets it from zero to non-zero), and then later within the same transaction deallocates the same cell (i.e., sets it back to zero), then the gas paid for the allocation is refunded to the transaction's gas payer at the end of the transaction.

edfelten:
This refunds gas charged for allocating a storage cell, if the same cell is deallocated before the end of the transaction. Refunds are given at the end of executing a top-level transaction, similar to when Ethereum gives storage refunds.

Fixes #622


rachel-bousfield (approved):
Looks great and passes all integration tests

PR #625, opened 2021-09-22
String ID refactor

RGates94:
This changes the way that StringIds are structured, so that they hold the string itself and the path of the file it is contained in.  This has the benefit of making each StringId unique across files and prevents them changing due to changes to order in the file.


src/link/mod.rs
```diff
@@ -164,10 +161,10 @@ impl Import {
         let name = name.to_string();
         let unique_id = Import::unique_id(&path, &name);
         Import {
+            id: Some(StringId::new(path.clone(), name.clone())),
```
rachel-bousfield, line 164: We probably should no longer let this field be optional now that IDs are always known

src/compile/ast.rs
```diff
@@ -169,7 +169,7 @@ impl Type {
 
         while let Type::Nominal(path, id, spec) = base_type.clone() {
             base_type = type_tree
-                .get(&(path.clone(), id))
+                .get(&(path.clone(), id.clone()))
```
rachel-bousfield, line 172: A flamegraph from earlier suggested that checking assignability is the plurality of our execution time. Clones, therefore, probably shouldn't happen here if possible. Moving to Nominal types being StringIds based would eliminate this clone.

src/compile/mod.rs
```diff
@@ -791,7 +761,7 @@ fn resolve_imports(
                 )
             })?;
 
-            let string_id = match origin_module.string_table.get_if_exists(&import.name) {
+            let string_id = match import.id.clone() {
                 Some(string_id) => string_id,
                 None => {
```
rachel-bousfield, line 766: This is never None

src/compile/mod.rs
```diff
@@ -1018,12 +970,32 @@ fn codegen_programs(
...
 
+    let main_func_index = if let Some(main_module) = typechecked_modules.get(0) {
+        main_module
+            .checked_funcs
+            .iter()
+            .position(|(id, _func)| id.id == "main")
+            .ok_or_else(|| {
+                CompileError::new(
+                    format!("Compile error"),
+                    format!("No \"main\" function found in \"{}\"", main_module.name),
```
rachel-bousfield: New error messages would be best with colors instead of `""`. Also `CompileError` now takes `str` so you don't need the `format!`. Don't worry about the others in the codebase since I've changed them in another PR, but new error messages should use `str`.

src/compile/mod.rs
```diff
@@ -1051,52 +1023,50 @@ fn codegen_programs(
...
                 func_labels.clone(),
-                module.string_table.clone(),
                 global_vars.clone(),
                 module.name.clone(),
                 module.path.clone(),
             ));
         }
     }
 
+    work_list.swap(0, main_func_index);
```
rachel-bousfield: Instead of checking and then re-ordering the work list so that `main` is the first function, it may be simpler to DFS from `main` during `link()`

src/compile/mod.rs
```diff
@@ -1115,7 +1085,7 @@ fn codegen_programs(
 
     let mut globals: Vec<_> = globals.into_iter().map(|x| x.1).collect();
     globals.push(GlobalVar::new(
-        usize::MAX,
+        StringId::new(vec![], String::new()),
```
rachel-bousfield: I like the idea of having `meta` be the source file this comes from, just like how we do `entry_point`. It's not needed, but for consistency's sake with the linker, it'd be cool to have stuff that comes from no module be the `meta` module.

src/compile/typecheck.rs
```diff
@@ -662,8 +663,8 @@ fn flowcheck_liveliness(
                 },
                 TypeCheckedNode::Expression(expr) => match &mut expr.kind {
                     TypeCheckedExprKind::Loop(..) => true,
-                    TypeCheckedExprKind::DotRef(_, id, ..)
-                    | TypeCheckedExprKind::LocalVariableRef(id, ..) => {
+                    //TODO: Dot refs are broken
```
rachel-bousfield: I got rid of the DotRef AST node in `#basic-blocks` so this is fine :)

rachel-bousfield (changes requested):
I'm really excited to see the removal of the `string_table` and the global consistency of `StringId`s. There's a few changes we should make to make this even better. Along with the comments I left in various places, we'll want to make the `Nominal` type no longer include a path. Right now it's (`path`, (`path`, `name`), `Type`), meaning the first `path` is redundant. This has consistency implications & makes certain operations more expensive due to clones. As it is right now, the compiler is now ~30% slower than when `StringId`s were integers and this may help with that. It's probably worth thinking through each operation again and seeing if there's a clone-free way to do it.

src/link/mod.rs
RGates94: I felt this would be better to put in a follow up PR to keep the scale of changes down. In general I find it is easier to merge multiple smaller changes than a single very large change. I recognize there are some benefits to changing it here, but I am not convinced they necessarily are worth the additional size of the PR.

src/compile/ast.rs
RGates94: I agree with this change, but I believe that this might be better as a follow up PR per the logic above.

src/compile/mod.rs
RGates94: I think it would be better to not make this change to prevent the possibility of errors resulting from a `meta.mini` file.

src/compile/mod.rs
RGates94: I think it would be better to put this in a follow up PR as per the logic above.

src/compile/typecheck.rs
RGates94: Good catch, I forgot to remove this.

src/compile/mod.rs
RGates94: While I think I see how to do this, it is not clear to me how it would simplify the implementation. If you have a particular approach in mind would do this for this operation, I would appreciate if you were to implement it.

rachel-bousfield:
As requested, I've simplified how we use `main` so that we don't require worklist swapping code now that `StringId`s are unordered.
```rs
let main = graph.node_indices().find(|node| graph[*node].name == "main")
```
The way it works is that, instead of DFS-ing from the first node in the graph, we DFS from main.

PR #626, opened 2021-09-23
Upgrade 45

rachel-bousfield:
Upgrade ArbOS to version 45 via the following PRs
- #624 

This upgrade is unusual in that we must use it to make `mainnet` and `rinkeby` consistent again. `rinkeby-upgrade-45` is Rinkeby's mirror version, different only in its `arbos_before.mexe`
- #627


edfelten (dismissed):
Looks good.

hkalodner (dismissed):
LGTM

edfelten (approved):
Re-approving.

PR #628, opened 2021-09-24
Emulator tuple hash

rachel-bousfield:
Makes hashing of integer-carrying tuples consistent with the C++ emulator.


edfelten (approved):
Looks good.

PR #629, opened 2021-09-29
Ethcall overrides

tsahee:
This implements and adds tests for the overrides required by geth-style ethcall overrides.
https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set


arb_os/accounts.mini
```diff
@@ -362,6 +362,28 @@ public func account_setBaseTxFeeL1Gas(account: Account, fee: uint) -> Account {
...
+
+public func account_setStorage(account: Account, storage: StorageMap) -> Account {
+    return account with {
+        contractInfo: Some(
+            xif let Some(contractInfo) = account.contractInfo {
+                contractInfo with {
+                    storage: storage
+                }
+            } else {
+                struct {
```
edfelten, line 375: In the `else` case, it seems better to return a failure rather than creating an invalid contract. We might make `account_setStorage` return `option<Account>`, and then return `None` in the `else` case here.

edfelten (changes requested):
One suggestion for handling an error case.

arb_os/accounts.mini
tsahee: I improved the else case to behave like the revert precompile, and also treated "store" for a new contract.
We could get a long list of overrides, with duplicate entries etc.. it's possible that code override will follow storage override for a new contract, and possible that we'll get a storage override that will eventually be irrelevant for the call. I think it's better not to fail unless we need to.
If call ends up calling contract that has no code - it should fail. I think revert_precompile behavior makes sense..

arb_os/accounts.mini
```diff
@@ -362,6 +363,28 @@ public func account_setBaseTxFeeL1Gas(account: Account, fee: uint) -> Account {
...
+        contractInfo: Some(
+            xif let Some(contractInfo) = account.contractInfo {
+                contractInfo with {
+                    storage: storage
+                }
+            } else {
+                let fakeCode = bytearray_new(0);
+                fakeCode = bytearray_setByte(fakeCode, 0, 0xfe);
+                struct {
+                    contractStateVersion: 1,
```
edfelten: I would use `contractStateVersion: const:: ContractStateVersion_current,` here because that is future-proof if we change the contract state version.

arb_os/accounts.mini
```diff
@@ -362,6 +363,28 @@ public func account_setBaseTxFeeL1Gas(account: Account, fee: uint) -> Account {
...
+                    code: fakeCode,
+                    evmJumpTable: newmap<uint, view write func()>,
+                    startCodePoint: makeTrampoline(revertPrecompile),
```
edfelten: Instead of creating a revert precompile, you can use an existing function by adding a use statement for `std::codeSegment::getDummyAcceptContractCode` then putting `startCodePoint: getDummyAccptContractCode(),` here.

edfelten (changes requested):
Two suggestions to simplify the dummy contract creation.

edfelten (approved):
Looks good.  Thanks for the fixes.

PR #631, opened 2021-10-10
Graph reduce

rachel-bousfield:
This PR Re-expresses basic blocks into value graphs, enabling connectivity-aware optimization passes. Included are the following optimization strategies:
- Constant folding
- Static simulation of pure function calls
- Detuplification via `swaps` & `pops`
- Local variable elision & lifetime analysis

Additionally, this PR
- Adds the `sensitive` qualifier for funcs that violate the function call ABI
- Adds the `δ-drop` opcode for notating the lifetime of a local
- Adds the `-O<usize>` flag for specifying the optimization level
- Adds stack transformation mechanisms for arbitrarily mutating the stack
- Adds a `Pop(n)` virtual opcode for removing items deep in the stack
- Adds console tooling for side-by-side comparisons of code segments & graphs
- Updates the emulator to detect divergence from the C++ one


src/optimize/mod.rs
```diff
@@ -45,6 +57,14 @@ pub enum BasicEdge {
 pub struct BasicGraph {
     /// Basic blocks and the edges that connect them
     graph: StableGraph<BasicBlock, BasicEdge>,
+    /// The Func this graph represents
+    func_id: LabelId,
+    /// The output node of the graph
```
edfelten, line 62: Is this comment correct?

edfelten (changes requested):
Just one small comment.

src/optimize/mod.rs
rachel-bousfield: Yes, a Basic Graph is always derived from a function's source code. The basic blocks are arranged in ways that take advantage of this -- for example, return statements link to the output node

edfelten (approved):

PR #639, opened 2021-11-05
Upgrade 49

rachel-bousfield:
This upgrade includes the following PR, which makes gas estimation's behavior consistent in the case where the gas price is zero
- https://github.com/OffchainLabs/arb-os/pull/638


hkalodner (approved):
LGTM

PR #640, opened 2021-11-08
Upgrade 49

edfelten:
Fix a bug in the modexp precompile, which was caused by a bug in the biguint library.


rachel-bousfield (approved):
Looks good and passes integration tests

PR #641, opened 2021-11-09
Change the default values of arrays and fixed arrays

RGates94:
This changes the default values of arrays and fixed arrays to improve type safety.


rachel-bousfield (approved):
I've added a test for this and it passes

edfelten (approved):
Looks good.

PR #642, opened 2021-11-11
Modexp Improvements & Test

rachel-bousfield:
Adds a test for `modexp` as well as an emulator fix for `SetBuffer64`


minitests/modexp.mini
```diff
@@ -0,0 +1,151 @@
...
+
+    let rand_bytes = view write closure(lower: uint, upper: uint) -> []uint {
+        let bytes = bytearray_new(0);
+        let i = 0;
+        let length = lower + rand(upper - lower);
+        while i < length {
+            bytes = bytearray_set256(bytes, i*32, bytearray_get256(entropy, rand(1)));
+            i = i + 1;
+        }
+        biguint_fromBytes(bytearray_extract(bytes, 0, 1))
```
edfelten, line 110: It looks like this will always produce a biguint that is less than 256, because the bytearray_extract will return a bytearray of length 1.

minitests/modexp.mini
```diff
@@ -0,0 +1,151 @@
...
+        let case = unsafecast<([]uint, []uint, []uint, []uint)>(case);
+        let (base, exponent, modulus, expected) = case;
+
+        let result = biguint_modexp(base, exponent, modulus)?;
+        if !biguint_equal(result, expected) {
+            return Some(("failed on test #".1, (queue_size(q) + 1), "expected:", expected, "but have:", result));
+        }
+    }
+
+    entropy = bytearray_new(0);
```
edfelten, line 93: It's better to use the functionality in `std::random` to generate pseudorandom numbers.  That might need some additions, like a `random_bytes` function, but it's better to add that to the standard library so that other code can benefit.

edfelten (changes requested):
A couple of issues with the new testing code.

PR #643, opened 2021-12-01
Don't charge for estimation when gas price is 0

rachel-bousfield:
This PR permits estimation when the gas price is 0, even for accounts that cannot pay.
In practice, this allows an unfunded account to estimate transaction costs it'll do in the future.


edfelten (approved):
Looks good. Thanks for tracking this down.

PR #644, opened 2021-12-02
Upgrade 50

rachel-bousfield:
Upgrade's ArbOS to the 50th version, which will include the following PRs
- #643 
- #642 
- #645

In summary, this upgrade
- Allows estimation for unfunded accounts when the gas price is 0
- Makes modexp more performant


edfelten (dismissed):
Looks good.  All of these changes were also previously reviewed before being merged into here.

edfelten (approved):
Re-approving after merging.

PR #652, opened 2021-12-17
Change getCurrentTxL1GasFees API

edfelten:
Change API to return a single L1 gas fees number, no longer breaking it down into two categories.  This will be more future-proof as we move to Nitro,


rachel-bousfield (approved):
LGTM, merging

PR #653, opened 2021-12-17
Fix GasPrice Zero Estimation

rachel-bousfield:
Fixes gas estimation for contract calls & deploys for unfunded accounts when the gas price is zero


edfelten (approved):
Looks good.

PR #656, opened 2022-01-10
Fix passing reference

eigmax:


rachel-bousfield:
Adding a leading `&` doesn't change the hash, and we want to move the value. Thanks for the PR though!

eigmax:
Why should we move?  it's deconstructed after the block is executed automatically.  What's your rust version?    I use rustc 1.53.0-nightly and get error: 

image

eigmax:
@RachelFranks

rachel-bousfield:
Your rust is out of date, please update to the latest release

PR #658, opened 2022-01-25
Remove tx gas limit for off-chain calls

edfelten:
This removes the transaction gas limit for "admin mode" calls.  We assume that node operators will impose whatever limit they want, if they want one, by using command-line options in the node software.

This also limits the size of deployed contract (EVM) code to 48 KB.  There are exceptions for (1) previously deployed contracts, and (2) new contracts whose deployed code is identical to a previously deployed contract.  The size becomes a chain parameter, so it can be adjusted by the chain owner as necessary.


rachel-bousfield (approved):
Looks great! I've done the full testing suite and it passed everything after a `make push`

PR #659, opened 2022-01-25
Upgrade 53

rachel-bousfield:
Upgrades ArbOS to the 53rd version via the following PR
- https://github.com/OffchainLabs/arb-os/pull/658

In summary, this upgrade
- Sets a configurable upper bound on contract code sizes
- Fixes an unnecessary gas check in admin mode


edfelten (approved):
Looks good.

PR #663, opened 2022-01-26
Use momentum in l2 gas pricing

edfelten:
This changes the L2 gas price adjustment algorithm, as follows:
* adds a "momentum" term that uses a rateEstimator abstraction to estimate the rate at which gas is being consumed, in a decaying weighted average with 60-second time constant,
* sets a target for the gas pool fullness (previously, this was implicitly 50% of the gas pool max),
* rewrites the L2 price adjustment algorithm, to be based on two factors: the ratio of gas usage rate estimate to chain speed limit, and the ratio of the gas pool fullness percentage to the target fullness
* uses an approximation to the exp() function to compute how much to change the price, with the rate of change clamped (as in the previous implementation) so that it does not exceed the max rate of change possible on average under Ethereum's EIP-1559

The estimator's time constant (default 60 seconds) and target gas pool fullness (default 80%) are added as chain parameters.

For convenience, this also includes an unrelated change allowing the programmer to specify the order in which the Mini compiler packs globals into the AVM Register. This allows a modest improvement in ArbOS performance.


arb_os/gasAccounting.mini
```diff
@@ -1071,6 +1096,12 @@ view throw func updateCongestionState(state: CongestionState, toBlockNum: uint,
     let gasPoolMax = chainParams_getOrDie(const::Atom_Param_GasPoolMax);
     let speedLimit = chainParams_getOrDie(const::Atom_Param_SpeedLimitPerSecond);
 
+    let rateEstimator = rateEstimator_advanceTime(state.gasUsageRateEstimator, toTimestamp);
+    let rateFactor = rateEstimator_get(rateEstimator) * gasPoolMax / (2 *speedLimit);
```
rachel-bousfield: I think we should remove the `2` here

Currently we have the following
```rs
price' = price * (120 * pool_max + factor - pool) / (120 * pool_max)
where factor = min(pool_max, estimate * pool_max / (2 * speed))
```
Suppose the price is rising
```rs
price' = price * (120 * pool_max + factor - pool) / (120 * pool_max)
price' > price 
=> factor > pool 
=> min(pool_max, estimate * pool_max / (2 * speed)) > pool
=> estimate * pool_max / (2 * speed) > pool  // pool_max exceeds pool
```
As we approach the steady state, `estimate` will approximate the speed limit, so
```rs
speed * pool_max / (2 * speed) > pool
pool_max / 2 > pool
=> The pool will be half full at equilibrium
```

rachel-bousfield (changes requested):
We'll need to make a minor change to bring equilibrium where we want it -- otherwise looks good (the estimator makes sense & the overall structure is solid)

arb_os/gasAccounting.mini
rachel-bousfield: Having discussed and clarified the 50% is intended to be a short-term buffer, let's finalize a constant (perhaps still 2) and then I'll approve this :)

rachel-bousfield (approved):
Looks great!

arb_os/gasAccounting.mini
```diff
@@ -195,10 +208,40 @@ view write throw func setupParameters(isInit: bool) {
...
+                    gasPoolMax / 2
+                } else {
+                    gasPoolMax * x / 100
+                }
+            } else {
+                gasPoolMax / 2
+            }
+        },
+        gasPoolWeight: {
+            if let Some(x) = chainParams_get(const::Atom_Param_GasPoolFullnessTargetPercent) {
```
hkalodner, line 235: Shouldn't this be `Atom_Param_PoolFullnessRatioWeightBasisPoints`

arb_os/gasAccounting.mini
```diff
@@ -195,10 +208,40 @@ view write throw func setupParameters(isInit: bool) {
...
+                if x > 10000 {
+                    10000
+                } else {
+                    x
+                }
+            } else {
+                const::Atom_Param_PoolFullnessRatioWeightBasisPoints
```
hkalodner, line 242: Shouldn't this be `Default_Param_PoolFullnessRatioWeightBasisPoints`?

arb_os/gasAccounting.mini
```diff
@@ -195,10 +208,40 @@ view write throw func setupParameters(isInit: bool) {
...
+            if let Some(x) = chainParams_get(const::Atom_Param_GasPoolFullnessTargetPercent) {
+                if x > 100 {
+                    gasPoolMax
+                } else if x < 50 {
+                    gasPoolMax / 2
+                } else {
+                    gasPoolMax * x / 100
+                }
+            } else {
+                gasPoolMax / 2
```
hkalodner, line 231: Maybe better to default to `Default_Param_GasPoolFullnessTargetPercent` instead like below

arb_os/gasAccounting.mini
```diff
@@ -680,6 +732,10 @@ public view write throw func gasAccounting_resumeTxCharges(allPurposeGas: uint,
             let osGasUsed = ((~0) - getGas()) / gasAccountingInfo.pricingParams.arbGasDivisor;
 
             set gasAccountingInfo.congestionState.gasPool = safeSubInt(gasAccountingInfo.congestionState.gasPool, int(osGasUsed));
+            set gasAccountingInfo.congestionState.gasUsageRateEstimator = rateEstimator_addUsage(
+                gasAccountingInfo.congestionState.gasUsageRateEstimator,
+                osGasUsed,
+            );
```
hkalodner, line 738: Maybe safer to combine these into a single function if they're always done in combination?

PR #664, opened 2022-02-02
Revert data from constructors

edfelten:
Fix bug that prevented reporting revert data from constructors that reverted, and add a test of this.


src/minitests/mod.rs
```diff
@@ -1480,6 +1480,11 @@ fn test_reverting_payable_constructor() {
         if receipt.clone().unwrap().succeeded() {
             panic!("unexpected success deploying PRConstructor contract");
         }
+        let error = Value::Buffer(Buffer::from_bytes(
```
rachel-bousfield, line 1483: Note: I've made a slight tweak to check for the phrase `revert message`

rachel-bousfield (approved):
Looks great

PR #668, opened 2022-02-08
Upgrade 54 Patches

rachel-bousfield:
Fixes a chain params bug and improves style


hkalodner (approved):
LGTM

arb_os/gasAccounting.mini
```diff
@@ -228,18 +228,18 @@ view throw func initCongestionState(isInit: bool) -> CongestionState {
                     gasPoolMax * x / 100
                 }
             } else {
-                gasPoolMax / 2
+                const::Default_Param_GasPoolFullnessTargetPercent
```
edfelten: This needs to be multiplied by `gasPoolMax` then divided by 100.

edfelten (changes requested):
One bug to fix.

arb_os/gasAccounting.mini
rachel-bousfield: Nice catch, fixed

edfelten (approved):
Looks good.

PR #669, opened 2022-02-08
Owners induce Admin Mode

rachel-bousfield:
This PR makes it so that when a caller is a chain owner, the tx will run in admin mode & skip congestion state updates


edfelten (approved):
This looks good.  Noting that this removes the application of the `allowedSenders` list, which we used to limit who could send txs early in the history of the chain.

rachel-bousfield:
Superseded by #674 
- #674

PR #671, opened 2022-02-15
Gas Accounting Refactors

rachel-bousfield:
This PR refactors gas accounting by
- Removing `adminMode` and adding `doNotCharge`, which has the same effect but includes a couple more cases
- Separating out the nonce check from `gasAccounting_prepareTx`
- Moving receipt emission to `handleArrivingTx
- Various miscellaneous simplifications


rachel-bousfield:
Superseded by #674 
- #674

PR #672, opened 2022-02-15
Upgrade 55

rachel-bousfield:
Upgrades ArbOS to the 55th version via the following PRs
- #668
- #663
- #674
- #675 

In summary, this upgrade includes
- A new congestion model, making prices both more tunable and less jittery when demand spikes
- A state correction pathway should administrative fixes be necessary
- A minor bug fix for `setbuffer8` and `setbuffer64`


edfelten (approved):
Approved per previous reviews.

PR #674, opened 2022-02-18
State Correction Txes

rachel-bousfield:
Though not intended to be used, this PR adds the ability for the bridge to emit a `stateCorrection` transaction in case of ArbOS state corruption. This safety mechanism allows for arbitrary code execution by address 0.


edfelten (approved):
Looks good.

PR #675, opened 2022-02-26
Truncate inputs to setbuffer8/64 in bytearray

edfelten:
This updates `bytearray_setByte` and `bytearray_set64` to truncate the values passed to `setbuffer8` and `setbuffer64`.


rachel-bousfield (approved):
Looks good and passes integration tests

PR #676, opened 2022-03-04
Simplify contract compilation setup

fredlacs:
This simplification (and control over the path where artifacts are generated) allow us to have better cross platform support in tooling that consumes the precompiles package


rachel-bousfield (approved):
Looks great!

PR #678, opened 2022-03-07
Deploy rlp fix

edfelten:
This tweaks how addresses are generated for contracts deployed via txs to null destination address.  This should fix an unexpected incompatibility for txs from senders whose addresses have leading zeroes.


rachel-bousfield (approved):
Looks great!

PR #682, opened 2022-04-20
Round up to multiple of 32 bytes in bytes ABI encoding in EVM state serialization

PlasmaPower:


rachel-bousfield:
State export moved to Nitro

PR #684, opened 2022-05-27
Add L1MessageType_ShutdownForNitro

PlasmaPower:
This triggers the needed actions (that I know of) to shutdown ArbOS classic and prepare for Nitro:
- Turn on the whitelist again and set it to owners only (just in case - there should be no way to submit messages at this point)
- Send any remaining outbox entries to L1
- Signal to the node that classic ArbOS is shut down by running a sideload with a block number of `1 << 128`, which will manifest in the machine's `l2BlockNumber` being that, which can easily be identified in any place it needs to be


edfelten (approved):
Looks good.

fredlacs:
We currently have a way of detecting in the L1 that this message was read, but not that it was successfully processed.

Discussed out-of-band the possibility of maybe adding a final send with a magic value that we can detect.
We could slightly change the behaviour of the [outbox](https://github.com/OffchainLabs/arbitrum/blob/1b1c05cdcadfba5271ae62cfb0b88f787381f406/packages/arb-bridge-eth/contracts/bridge/Outbox.sol#L113-L114) to make this easier.

edfelten (approved):
Looks good.

